PR #17, opened 2024-02-26
add check_usb_hub_connected to utils

hhatpsi:


src/pnpq/devices/utils.py
```diff
@@ -0,0 +1,28 @@
+from serial.tools.list_ports import comports as list_comports
+import logging
+
+logger = logging.getLogger("utils")
+
+USB_HUB_iVENDOR = "2109"
```
zigen, line 6: `USB_HUB_VENDOR_ID` is fine. I don't want to put `i` in the constant name.

src/pnpq/devices/utils.py
```diff
@@ -0,0 +1,28 @@
...
+USB_HUB_iPROFUCT = ["0817", "2817"]
```
zigen, line 7: the number of available usb hubs will grow. so, we should check by the tuple of vid and pid.
then the available usb hubs are defined like this
```py
AVAILABLE_USB_HUBS: list[tuple[str, str]] = [
  ("2109", "0817") # some usb hub name
  ("2109", "2817") # another usb hub name
]
```

src/pnpq/devices/waveplate_thorlabs_kb10crm.py
```diff
@@ -61,7 +64,9 @@ def __init__(
         self.home_timeout = 20
         self.max_channel = 1
         self.auto_update = False
+        self.hub = check_usb_hub_connected()
```
zigen, line 67: should be `self.hub_connected`. `self.hub` looks this represents a usb hub instance.

src/pnpq/devices/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    logger.debug(f"get_available_port(serial_number: {device_serial_number})")
+    available_ports = list_comports()
+    for port in available_ports:
+        if port.serial_number == device_serial_number:
+            logger.debug(f"port found: {port.device}")
+            return port.device
+    return None
+
+
+def check_usb_hub_connected() -> bool | None:
```
zigen, line 20: this returns only bool value. this must be `-> bool`

src/pnpq/utils.py
```diff
@@ -8,7 +9,23 @@ def get_available_port(device_serial_number: str) -> str | None:
...
     available_ports = list_comports()
     for port in available_ports:
+        logger.debug(f"port: {port.device}")
         if port.serial_number == device_serial_number:
             logger.debug(f"port found: {port.device}")
             return port.device
     return None
+
+
+def check_usb_hub_connected() -> bool | None:
```
zigen, line 19: this also always return boolean value.

zigen (changes requested):
please add pyproject.toml and requirements*.lock

src/pnpq/devices/utils.py
hhatpsi: AVAILABLE_USB_HUBS was defined to be a list of (vendor, product) pairs.

src/pnpq/utils.py
hhatpsi: method annotation was modified.

hhatpsi (commented):
remove os-dependent code for detecting usb hub and redefine check_usb_hub_connected method based on usb library.

zigen (approved):

Issue #21, opened 2024-04-18
Stub Mode for Optical Switch


zigen:
In order to aid future integration testing and help learn more about how our hardware works, we are building "stub" versions of devices. These do not connect to real hardware, but instead mimic or mock the behavior of those devices in code.

Later on, this will help us define common interfaces for devices, instead of the device-specific interfaces that we have now. It may also help us find better ways to refactor our existing code.

## Acceptance criteria

* Stub switch has same public interface as [the Thorlabs switch](https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/devices/switch_thorlabs_osw1310E.py)
* Stub switch behavior reflects happy-path behavior for the Thorlabs switch. In other words, the stub replicates the same behavior as the real switch when the switch does not produce any errors.
* All methods have type hints for input variables and return values.
* All public methods have basic docstring documentation explaining the method's behavior clearly. The class may also have docstring documentation.
* Any unclear variables or code is adequately commented.
* `pytest` unit tests are written for all public methods, with meaningful inputs

auspicacious:
@master-of-ppap Welcome to Thunderdome. I suggest looking at https://github.com/moonshot-nagayama-pj/PnPQ/pull/23 for an idea of what the final product will look like.

@TomShen1234 Did I miss anything in the acceptance criteria? I am planning on writing a more general "definition of done" in the future.

TomShen1234:
Looks fine to me right now.

master-of-ppap:
> @master-of-ppap Welcome to Thunderdome. I suggest looking at #23 for an idea of what the final product will look like.
Thank you! Is there any documentation for what the optical switch does/is supposed to do?

auspicacious:
Is there any documentation for what the optical switch does/is supposed to do?

@hhatpsi and @Marie2222 would be good people to ask about that.

auspicacious:
Initial stub is done, so I'm marking this as done. After refactoring in https://github.com/moonshot-nagayama-pj/PnPQ/issues/27 we will have more work to do here.

PR #23, opened 2024-06-26
Stub waveplates

TomShen1234:
Implementation for #19

Features:

- Simulated position check
- Simulated serial check (connect check only, does not create any serial connection, checks to see if connect() is called)
- Functions are in sync with Thorlabs kb10crm
- Enable/disable channels (currently the motor for stub device is on channel 1, we may need to verify If that's the actual behavior)

Next steps:

- Simulate errors
- Tentative: Return fake replies from function
- Integrate into testbed software
- Unified logs


src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,235 @@
...
+        self.__ensure_port_open()
+        self.__ensure_valid_degree(degree)
+
+        if not self.__stub_check_channel(1):
+            # Do nothing if channel is not enabled
+            return
+
+        self.logger.info(f"Stub Waveplate Rotate to {degree}")
+        # Calculate number of steps to move
+        movePosition = degree * self.resolution
```
auspicacious, line 145: This variable name doesn't follow [PEP 8](https://peps.python.org/pep-0008/#function-and-variable-names) conventions.

We aren't doing automatic linting in this project yet, which would catch errors like that, but we will do it in the future.

```suggestion
        move_position = degree * self.resolution
```

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,235 @@
...
+        self.current_position = movePosition
+        # Delay to simulate rotation (for now: v=1ms/deg)
+        time.sleep(abs(movePosition - self.current_position) / 1000)
+        # TODO: Return a fake reply from the device
+
+    def step_backward(self, steps: int):
+        self.__ensure_port_open()
+        self.__ensure_less_than_max_steps(steps)
+
+        self.logger.info(f"Stub Waveplate Step Backware {steps}")
```
auspicacious, line 156:
```suggestion
        self.logger.info(f"Stub Waveplate Step Backward {steps}")
```

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,235 @@
...
+            return
+
+        self.logger.info(f"Stub Waveplate Rotate to {degree}")
+        # Calculate number of steps to move
+        movePosition = degree * self.resolution
+        # Update current position
+        self.current_position = movePosition
+        # Delay to simulate rotation (for now: v=1ms/deg)
+        time.sleep(abs(movePosition - self.current_position) / 1000)
+        # TODO: Return a fake reply from the device
```
auspicacious, line 150: Will this work properly without a return value?

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,235 @@
...
+            # Do nothing if channel is not enabled
+            return
+
+        # Make steps negative
+        steps = -steps
+        self.current_position += steps
+        # Delay to simulate rotation (for now: v=1ms/deg)
+        time.sleep(abs(steps) / 1000)
+
+    def step_forward(self, steps: int):
```
auspicacious, line 168: I think that `step_forward` and `step_backward` can be implemented simply by calling `rotate` instead of copy/pasting `rotate`'s code?

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,235 @@
...
+        if not self.__stub_check_channel(1):
+            # Do nothing if channel is not enabled
+            return
+
+        self.current_position += steps
+        # Delay to simulate rotation (for now: v=1ms/deg)
+        time.sleep(abs(steps) / 1000)
+        # In the future, maybe we can return a fake reply from the device
+
+    def rotate_relative(self, degree: int | float):
```
auspicacious, line 183: Likewise, it might be easier to calculate the absolute rotate value and just call `rotate`.

auspicacious (commented):
I think we have something pretty close to a first usable version here. Many of my comments are about spelling errors, etc.

However, there are a couple of general considerations:

* All functions should define their return type.
* Adding some docstrings would be useful to explain concepts like connection, channel, and custom homing. Eventually, we can re-use some of that documentation 

There are also some long-term considerations that we don't need to solve right now:

* How can we use the context manager model to avoid checking for connections/active channels every time?
* How can we create a common waveplate interface class that inherits from this one?

src/pnpq/devices/waveplate_stub.py
TomShen1234: The original function seems to return values of type 'bytes'. It seems difficult to do the same for stub right now.

src/pnpq/devices/waveplate_stub.py
TomShen1234: The real device also seems to re-implement the rotating action. But it seems unnecessary, so let's not do that, and maybe we should fix the real one too.

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,235 @@
...
+
+        if not self.__stub_check_channel(1):
+            # Do nothing if channel is not enabled
+            return
+
+        self.current_position = 0
+        # Simulated delay for homing
+        # TODO: Delay calculation
+        time.sleep(1)
+        self.logger.info(f"Home Position: {self.current_position}")
```
auspicacious, line 88: It occurred to me that, when logging, it's better to use [lazy evaluation](https://realpython.com/python-f-strings/#lazy-evaluation-in-logging) (and I hope that there are static analyzers for Python like there are for Java that detect this sort of thing).

src/pnpq/devices/waveplate_stub.py
TomShen1234: I remembered why this is done like this. There seems to be an inconsistency between the functions. Some takes a degree value before converting to steps, and others uses steps directly. I wonder if this inconsistency is there on purpose. I'll do a manual conversion for now.

TomShen1234:
Summary of updates in the previous 2 commits:
- Spell error fixes
- Logs now uses lazy string interpolations
- Every advanced rotate function now calls the implementation of `rotate`
- Return types
- Simple docstrings

Further Actions still needed (they may be done in a separate issue):
- Most functions return `None`, it may be useful to return something useful once unified APIs are decided
- As stated above, we should have unified APIs for devices which all classes conforms to
- Log structure

TomShen1234:
It also came to be that we should name this `WaveplateStub` instead of `Waveplate` which has a name conflict with the (poorly named) real wave plate driver. 

Also, the init file now has an import.

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,217 @@
...
+        # Stub Serial Number
+        # TODO: Custom serial number from initializer
+        self.device_sn = "stubwaveplate"
+
+        self.resolution = 136533
+        self.max_steps = 136533
+        self.rotate_timeout = 10
+        self.home_timeout = 20
+        self.max_channel = 1
+        self.auto_update = False
```
auspicacious, line 33: 1. This isn't really the right way to define variables in Python, I think. The variables on 19-21 are static, and I don't think there's ever a value being assigned to them. The variables with the same names on lines 28-33 are actually totally separate. See [this StackOverflow discussion](https://stackoverflow.com/questions/9056957/correct-way-to-define-class-variables-in-python) for more information.
2. That being said, it would be a good idea to provide type information for the `self` variables.
3. Also, it would be a VERY good idea to add comments explaining what these variables actually mean and how they relate to each other. What is "resolution?" How do you calculate degrees from it? What are the timeouts?
4. It might be a good idea to implement a method that calculates and returns the current absolute position in degrees. This might follow the Python getter pattern.

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,217 @@
...
+        return self.resolution
+
+    def getpos(self) -> int | float:
+        """Get the current position of the device"""
+        self.__ensure_port_open()
+        self.logger.info("Stub Waveplate Get Position")
+        self.logger.info("Current Position: Steps: %s Degrees: %s", self.current_position, self.current_position / self.resolution)
+        return self.current_position
+
+    def rotate(self, degree: int | float) -> None:
```
auspicacious, line 141: This class stores its positional state in steps, not degrees. It therefore seems kind of backwards to me that `step_forward` and `step_backward` are calling `rotate`, instead of `rotate` calling a `step` function.

auspicacious (commented):
I think that it might be a good idea to do a little test-driven development here. Some of these errors might be easier to detect if you have an easy way to run the code you're writing.

src/pnpq/devices/waveplate_stub.py
TomShen1234: I agree with 1 - 3, but I think there may be an issue with 4 if we use Python's getter because on the real device, `getpos()` could throw errors. Maybe it will just be in the form of another function which calls the `getpos()` function.

src/pnpq/devices/waveplate_stub.py
TomShen1234: Currently, this is based off of the physical device driver. If we do it here, should we also make the same set of changes on the device driver?

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,217 @@
...
+
+        if not self.relative_home:
+            self.logger.error("Custom Home not set")
+            raise WavePlateCustomRotateError("Waveplate({self}) relative_home not set")
+
+        if not self.__stub_check_channel(1):
+            # Do nothing if channel is not enabled
+            return
+
+        self.rotate(degree + self.relative_home)
```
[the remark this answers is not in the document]
TomShen1234: Interestingly enough, the real device driver also does the checks in the same way...
I do think it should be changed though, it may be an error.

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,217 @@
...
+        self.__ensure_port_open()
+        self.__ensure_valid_degree(degree)
+
+        if not self.__stub_check_channel(1):
+            # Do nothing if channel is not enabled
+            return
+
+        self.logger.info("Stub Waveplate Rotate to %s", degree)
+        # Calculate number of steps to move
+        move_position = degree * self.resolution
```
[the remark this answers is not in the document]
TomShen1234: I'll check up on the `max_step` and `resolution` part, this section was originally copied over from the real device driver, and it also has the same value.

src/pnpq/devices/waveplate_stub.py
auspicacious: > here may be an issue with 4 if we use Python's getter because on the real device, `getpos()` could throw errors

We might be talking about different patterns here. I'm thinking about a function.

src/pnpq/devices/waveplate_stub.py
auspicacious: > Currently, this is based off of the physical device driver. If we do it here, should we also make the same set of changes on the device driver?

I'm not really sure what this means.

We want all of our waveplate drivers to have the same API -- the same interface. But the internal implementation can be different.

I don't think I'm asking for a change to the API here. An API change would require changing the physical device driver as well.

The internal implementation of the physical device driver does not work the same way as this, either.

TomShen1234:
Many changes were made (hence the large number of commits).
Everything I marked with a thumbs-up above were implemented. As well as unit test (using `pytest`) were added.

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def __init__(self):
+        # Stub Serial Number
+        # TODO: Custom serial number from initializer
+        self.device_sn = "stubwaveplate"
+
+        # Resolution of the device in steps per degree
+        self.resolution: int = 136533
+        # Maximum steps the device can move (360 degrees)
+        self.max_steps: int = 136533*360
```
TomShen1234, line 28: I'll add a commit here myself. In the real wave plate driver `max_step` seems to equal `resolution`, which means in the check for `step` function the wave plate can only rotate 1 degree. Is this intended? Right now I temporarily multiplied it by 360 degrees for it to pass the tests. 
Let me know if this is the wrong intended behavior. 
@zigen @hhatpsi

tests/waveplate_stub_test.py
```diff
@@ -0,0 +1,108 @@
...
+    assert len(wp.enabled_channels) == 1
+
+def test_step_forward_backward():
+    wp = WaveplateStub()
+    wp.connect()
+    wp.rotate(90)
+    pos = wp.getpos()
+    wp.step_forward(10)
+    newPos = wp.getpos()
+    assert newPos == pos + 10
```
auspicacious, line 79: Remember, camelCase variable names are not idiomatic Python. Should be `new_pos`.

```suggestion
    original_position = wp.getpos()
    wp.step_forward(10)
    assert wp.getPos() == original_position + 10
```

tests/waveplate_stub_test.py
```diff
@@ -0,0 +1,108 @@
...
+    wp.step_backward(10)
+    newPos = wp.getpos()
+    assert newPos == pos
```
auspicacious, line 82: `step_backward` should be tested in a separate test.

Unit tests should focus on the smallest unit of work possible.

tests/waveplate_stub_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_invalid_steps():
+    wp = WaveplateStub()
+    wp.connect()
+    with pytest.raises(WaveplateInvalidStepsError):
+        wp.step_backward(1)
+        wp.step_forward(wp.max_steps + 1)
```
auspicacious, line 89: `step_backward` and `step_forward` need to be in separate unit tests.

As it is, this will only test `step_backward`. `step_forward` will never be called.

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,219 @@
...
+        new_steps = self.current_position - steps
+        self.__set_steps(new_steps)
+
+    def step_forward(self, steps: int) -> None:
+        """Step forward by a specified number of steps"""
+        # Get new position and set steps
+        new_steps = self.current_position + steps
+        self.__set_steps(new_steps)
+
+    def rotate_relative(self, degree: int | float) -> None:
```
auspicacious, line 180: So, as far as I know, type hints are [only used by static analysis tools, not the interpreter](https://docs.python.org/3/library/typing.html). If you are using VSCode and it's complaining about this line when only `float` is set instead of `int | float`, I would be curious to see the error.

I have realized that I need to do more research on using static analysis tools before I understand which is the "right" way to do this in Python. My intuition might be wrong.

tests/waveplate_stub_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_rotate_relative():
+    wp = WaveplateStub()
+    wp.connect()
+    wp.rotate(90)
+    pos = wp.getpos()
+    wp.rotate_relative(10)
+    newPos = wp.getpos()
+    assert newPos == pos + 10*wp.resolution
```
auspicacious, line 98: The repeated use of `* wp.resolution` in these tests is a sign that the `WaveplateStub` class needs a `get_position_by_degree` method.

(Ideally `getpos` should be refactored to `get_position_by_step`).

I think this should be a follow-up GitHub task.

tests/waveplate_stub_test.py
```diff
@@ -0,0 +1,108 @@
+from pnpq.devices.waveplate_stub import WaveplateStub
+from pnpq.errors import (
+    DeviceDisconnectedError,
+    WaveplateInvalidDegreeError,
+    WaveplateInvalidMotorChannelError,
+    WaveplateInvalidStepsError,
+)
+import pytest
+
+def test_access_without_connection():
```
auspicacious, line 10: Every method that should throw this error should be tested.

How do we do this quickly and easily?

Let's use a [parameterized test](https://docs.pytest.org/en/8.0.x/how-to/parametrize.html).

More generally, this pattern is called data-driven testing.

Using `eval` like in the pytest documentation is kind of scary (`eval` should be approached with extreme caution) but it's also the simplest way to accomplish this, so let's give it a try.

tests/waveplate_stub_test.py
```diff
@@ -0,0 +1,108 @@
...
+    wp = WaveplateStub()
+    with pytest.raises(DeviceDisconnectedError):
+        wp.identify()
+
+def test_connect():
```
auspicacious, line 15: This doesn't test `connect`, it tests `identify`.

tests/waveplate_stub_test.py
```diff
@@ -0,0 +1,108 @@
...
+    wp = WaveplateStub()
+    wp.connect()
+    wp.identify()
+
+def test_home():
```
auspicacious, line 20: Creating a WaveplateStub and calling `connect` in every test over and over again is really repetitive.

Instead, let's use a [test fixture](https://docs.pytest.org/en/8.0.x/how-to/fixtures.html). Let's create a fixture called `connected_waveplate` and pass that in instead.

auspicacious (commented):
There's more that could be said, and more tests that could be written, but I think that this is enough comments for now.

So, how do we complete this PR?

First, we need to satisfy this round of comments.

That means both writing more code, and creating some follow-up tasks for our GitHub task board. I think that we might need a new epic called "PnPQ refactoring."

Once that is done, I will take one more look. I am hoping that after this next round of comments, I can say "this is good enough" and we can merge this, assuming that we have written down further improvements in GitHub tasks.

tests/waveplate_stub_test.py
TomShen1234: In the stub device `connect` do not fail, but the goal of the test is to check whether the device works after calling `connect` (and test whether requested are rejected if `connect` is not called). `identify` happens to be the simplest function to test this, as in the stub it doesn't do anything (but fails if the device is not connected)

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,219 @@
...
+        self.max_steps: int = 136533*360
+        # Maximum channel the device can control
+        self.max_channel: int = 1
+        # Flag for auto updating device information
+        self.auto_update: bool = False
+
+        self.logger = logging.getLogger(f"{self}")
+
+        # Current Position of the device in steps
+        self.current_position: int = 0
```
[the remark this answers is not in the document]
TomShen1234: oops... my bad

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,219 @@
...
+        """Rotate the device to a specified absolute degree"""
+        self.__ensure_port_open()
+        self.__ensure_valid_degree(degree)
+
+        if not self.__stub_check_channel(1):
+            # Do nothing if channel is not enabled
+            return
+
+        self.logger.info("Stub Waveplate Rotate to %s", degree)
+        # Calculate number of steps to move (round to nearest integer)
```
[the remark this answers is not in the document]
TomShen1234: Good point, the original driver also truncates the number, so instead of changing it to round, I'll modify the comment.

tests/waveplate_stub_test.py
TomShen1234: I modified it so that it says `test_identify`

tests/waveplate_stub_test.py
TomShen1234: I wrote a function that does exactly this in the stub (called `get_degree`), and it's used somewhat in the internal implementation. I didn't use it in the test because it's not present in the real driver (for compatibility reasons).

src/pnpq/devices/waveplate_stub.py
TomShen1234: The actual driver had this too (albeit for `rotate`, there's no type hint for `rotate_relative`), we may need to refactor this.

TomShen1234:
Everything marked with a thumbs-up is fixed. I'm not sure if the `eval` test is done well enough, it works in the mean time, but I can see it could be improved someways (not sure what's the best though, it's left as a TODO).

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,227 @@
...
+        """Rotate the device to a specified absolute degree"""
+        self.__ensure_port_open()
+        self.__ensure_valid_degree(degree)
+
+        if not self.__stub_check_channel(1):
+            # Do nothing if channel is not enabled
+            return
+
+        self.logger.info("Stub Waveplate Rotate to %s", degree)
+        # Calculate number of steps to move (truncate to nearest integer)
```
auspicacious, line 161: The "nearest" is confusing.

```suggestion
        # Calculate number of steps to move (truncate to integer)
```

src/pnpq/devices/waveplate_stub.py
```diff
@@ -0,0 +1,227 @@
...
+        move_position = int(degree * self.resolution)
+        # Update current position
+        self.__set_steps(move_position)
+        # Delay to simulate rotation (for now: v=1ms/deg)
+        time.sleep(abs(move_position - self.current_position) / 1000)
+        # TODO: Return a fake reply from the device
+
+    def step_backward(self, steps: int) -> None:
+        """Step backward by a specified number of steps"""
+        self.__ensure_port_open()
```
auspicacious, line 171: I see the extra unit tests are already finding bugs!

tests/waveplate_stub_test.py
```diff
@@ -0,0 +1,94 @@
...
+    wp = WaveplateStub()
+    with pytest.raises(DeviceDisconnectedError):
+        # Since the device is not connected, all methods should raise DeviceDisconnectedError
+        # For simplicity, 1 is passed as argument for all methods, the value should be small enough to not raise any other errors
+        # TODO: Refactor this to be more flexible in the future
+        arg = ",".join(["1"]*argc)
+        eval(f"wp.{f}({arg})")
+
+@pytest.fixture
+def c_wp():
```
auspicacious, line 21: There's a reason I said `connected_waveplate` and not `c_wp`.

Short variable names are difficult to read and understand, especially when you are not familiar with the code. I am trying to slowly nudge our codebase towards using full English words.

For more on choosing variable names, see this summary of [_Clean Code_](https://dzone.com/articles/naming-conventions-from-uncle-bobs-clean-code-phil) by Robert Martin.

TomShen1234:
Here are the compilation of things that needs to be added to the Github log (probably unfinished):

- Most functions return None, it may be useful to return something useful once unified APIs are decided
- Unified logging
- Refactor some APIs in PnPQ to enable less redundant codes
- TODOs left in codes
 - Delay calculation for homing operation
 - Auto update
 - Returning fake replies from device logs (we can work on this alongside the first item in this list)
 - Support overflow rotation for custom rotate
 - A more flexible way to test connection (instead of passing `1` to all arguments, this works for connection because it's the first thing being checked, but if we want this kind of test for other parameters this structure may no longer work)

Some of these probably could be combined into a single task.

auspicacious (approved):
Thank you for sticking with this PR, I think that this has been a good process.

Thanks for summarizing the tasks that we can spin out of this. I'll take a little time soon and make up the GitHub tasks from that list.

TomShen1234:
An additional note: should the original issue be closed here, or should it wait for `testbed-node` to gain support for this stub device? This is not mentioned in the stub acceptance criteria in #21

PR #25, opened 2024-07-12
.gitignore update

TomShen1234:
Added `.envrc` into gitignore to ignore configuration files for direnv.

Additionally, the requirements lock file from `rye sync` is updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -185,7 +185,8 @@ cython_debug/
 .LSOverride
 
 # Icon must end with two \r
-Icon
+Icon
+
 
 # Thumbnails
 ._*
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -185,7 +185,8 @@ cython_debug/
 .LSOverride
 
 # Icon must end with two \r
-Icon
+Icon
```
zigen, line 188: keep \r\n

zigen (changes requested):

.gitignore
auspicacious: Editorconfig automatically removes whitespace at the end of lines, so it is difficult to keep this entry corrrect. We did a little research (read StackOverflow) and using a wildcard `?` instead is probably a reasonably safe way to do this.

```suggestion
Icon?
```

auspicacious (approved):

PR #33, opened 2024-07-19
Stub optical switch

master-of-ppap:
Implementation for #21 

Features:
- Simulated state switch between bar state and cross state of optical switch
- Functions have same interface with Thorlabs osw1310E

### DIFF
--- a/src/pnpq/devices/switch_stub.py
+++ b/src/pnpq/devices/switch_stub.py
@@ -1,0 +1,43 @@
+import logging
+
+from pnpq.errors import (
+    DeviceDisconnectedError,
+)
+
+class Switch:
+    """Stub Switch Device Class"""
+
+    def __init__(self):
+        # Stub Serial Number
+        # TODO: Custom serial number from initializer
+        self.device_sn: str = "stubswitch"
+
+        # Is connected to the device (used internally)
+        self.connected: bool = False
+
+        # State of the optical switch
+        # 1 is bar state, 2 is cross state
+        self.state: int = 1
+
+        # Logger for this class
+        self.logger: logging.Logger = logging.getLogger(f"{self}")
+
+    def __ensure_port_open(self) -> None:
+        if not self.connected:
+            self.logger.error("Device not connected")
+            raise DeviceDisconnectedError(f"{self} is disconnected")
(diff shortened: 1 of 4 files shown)


src/pnpq/devices/switch_stub.py
```diff
@@ -0,0 +1,43 @@
+import logging
+
+from pnpq.errors import (
+    DeviceDisconnectedError,
+)
+
+class Switch:
+    """Stub Switch Device Class"""
+
+    def __init__(self):
```
auspicacious, line 10: I missed this in the earlier PR, but, here, I started wondering, "what should the return type of `__init__` be?"

And it turns out that [the type annotations PEP explicitly says that the return type of `__init__` should be `None`](https://peps.python.org/pep-0484/#the-meaning-of-annotations).

@TomShen1234 I missed a couple of type annotations on the waveplate PR. If you or @master-of-ppap could add them, it would be appreciated.

```suggestion
    def __init__(self) -> None:
```

src/pnpq/devices/switch_stub.py
```diff
@@ -0,0 +1,43 @@
...
+        # Stub Serial Number
+        # TODO: Custom serial number from initializer
+        self.device_sn: str = "stubswitch"
+
+        # Is connected to the device (used internally)
+        self.connected: bool = False
+
+        # State of the optical switch
+        # 1 is bar state, 2 is cross state
```
auspicacious, line 19:
```suggestion
        # 1 is bar state, 2 is cross state. These are the numbers used to represent these states in the Thorlabs Optical Switch 1310E driver.
```

auspicacious (commented):
After these two changes, I think this will be good enough for now. I think that we will need to make some improvements to the real device driver before making any more improvements to the stub.

master-of-ppap:
I heard from optical switch specialist miaさん that the default state of the optical switch depends on the actual physical state that the switch itself is set in, so that could result in some discrepancies when simulating. I'm not too sure though. Haha.

auspicacious (approved):

TomShen1234:
I'll add them into the Stub Waveplate.

PR #36, opened 2024-07-25
Return type hint in stub waveplate's initializer

TomShen1234:
Adds types hint `None` to return type of `__init__` for stub wave plate as mentioned in #33


auspicacious (commented):
@TomShen1234 Check line 61 as well.

TomShen1234:
Interesting, not sure why I missed that last time. Fixed

auspicacious (approved):

Issue #35, opened 2024-07-23
implement OdlOzOptics.current_step 


zigen:
I got this error:

```
>>> odl.current_step()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: OdlOzOptics.current_step() takes 0 positional arguments but 1 was given
```

and there's no implementation. the expected behavior is that this function returns current step value.
https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/devices/odl_ozoptics_650ml.py#L50

Issue #27, opened 2024-07-17
PnPQ API refactoring


auspicacious:
- [x] https://github.com/moonshot-nagayama-pj/PnPQ/issues/29
- [x] https://github.com/moonshot-nagayama-pj/PnPQ/issues/28
- [x] https://github.com/moonshot-nagayama-pj/PnPQ/issues/86
- [x] https://github.com/moonshot-nagayama-pj/PnPQ/issues/92

master-of-ppap:
@auspicacious would it be possible to close this issue now? I think we’ve done everything.

auspicacious:
Closing in favor of continuation in https://github.com/moonshot-nagayama-pj/PnPQ/issues/118

PR #37, opened 2024-07-25
Ozoptics odl bug fix #34 and #35

### LINKED ISSUE #34: OdlOzOptics.get_step cannot handle error response from the device properly
```
>>> odl.get_step()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/app/testbed-node/.venv/lib/python3.12/site-packages/pnpq/devices/odl_ozoptics_650ml.py", line 117, in get_step
    return int(step)
           ^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'UNKNOWN\r\n'
```

hhatpsi:
- Rewrite the main method to fix the bugs reported in #34 and #35. 
- Adding basic logging capability for debugging. 
- Logical bug fix found in for home() command (the previous command was find_home not home())
- Replace  more appropriate name for home() and end() methods (instead of forward and reverse)
- Verify the position of the odl after moving

### DIFF
--- a/src/pnpq/devices/odl_ozoptics_650ml.py
+++ b/src/pnpq/devices/odl_ozoptics_650ml.py
@@ -5,7 +5,14 @@ import serial
 from serial import Serial
 from pnpq.devices.optical_delay_line import OpticalDelayLine
 import time, logging
-from pnpq.errors import DeviceDisconnectedError
+from pnpq.errors import (
+    DeviceDisconnectedError,
+    OdlMoveNotCompleted,
+    OdlHomeNotCompleted,
+    OdlGetPosNotCompleted,
+    OdlMoveOutofRangeError,
+)
+
 
 
 class OdlOzOptics(OpticalDelayLine):
@@ -42,9 +49,11 @@ class OdlOzOptics(OpticalDelayLine):
         self.command_terminate = "\r\n"
         self.max_move = 50
         self.min_move = 0
+        self.logger = logging.getLogger(f"{self}")
 
         try:
             self.conn.open()
+
         except:
             raise RuntimeError("Can not open OZ optic ODL device")
 
@@ -53,43 +62,73 @@ class OdlOzOptics(OpticalDelayLine):
             self.logger.error("disconnected")
             raise DeviceDisconnectedError(f"{self} is disconnected")
 
-    def connect(self):
-        if self.conn.is_open == 0:
+    def reconnect(self):
+        if self.conn.is_open:
+            self.logger.warning(f"Serial connectio to ODL({self}) is already open!")
+        else:
+            self.logger.info(f"Openning connection to ODL({self}) device!")
             try:
                 self.conn.open()
-            except Exception as err:
-                raise RuntimeError("Connection failed: " + str(err))
+            except:
+                raise DeviceDisconnectedError(f"{self} is disconnected")
 
     def move(self, dist: float):
+        self.logger.debug(f"ODL({self}): move command is received")
         self.__ensure_port_open()
 
         if dist > self.max_move or dist < self.min_move:
-            raise Exception("Invalid Move Parameter")
+            raise OdlMoveOutofRangeError(f"ODL({self}): Invalid Move Parameter:{dist}")
         else:
-            self.set_step(int(dist * self.resolution))
+            response = self.set_step(int(dist * self.resolution))
+            self.logger.debug(f"ODL({self}): move command complete: {response} verifying:")
+            step_position = self.get_step()
+            self.logger.debug(f"ODL({self}): final position after move:{step_position/self.resolution}")
 
     def set_step(self, value):
+        self.logger.debug(f"ODL({self}): set_step command is received")
         self.__ensure_port_open()
         cmd = "S" + str(value)
         response = self.serial_command(cmd)
         return response
 
-    def current_step():
-        pass
+    def get_step(self):
+        self.logger.debug(f"ODL({self}): get_step command is received")
+        self.__ensure_port_open()
+        cmd = "S?"
+        response = self.serial_command(cmd)
+        if (response.find("UNKNOWN")):
+            raise OdlGetPosNotCompleted(
+                f"Unknown position for {self}: run find_home()
+                first and then change or get the position"
+            )
+        step = response.split("Done")[0].split(":")[1]
+        self.logger.debug(f"ODL({self}): get_step:{step}:return{int(step)}")
+        return int(step)
 
-    def home(self):
+    def current_status(self):
+        self.logger.debug(f"ODL({self}): current_status query is received!")
+        self.__ensure_port_open()
+        cmd = "Q?"
+        response = self.serial_command(cmd)
+        self.logger.debug(f"ODL({self}) current_status:{response}")
+        return response
+
+    def find_home(self):
+        self.logger.debug(f"ODL({self}): find_home command is received!")
         self.__ensure_port_open()
         cmd = "FH"
         response = self.serial_command(cmd, retries=1000)
         return response
 
     def get_serial(self):
+        self.logger.debug(f"ODL({self}): get_serial command is received!")
         self.__ensure_port_open()
         cmd = "V2"
         response = self.serial_command(cmd)
         return response.split("Done")[0].split("\r\n")[1]
 
     def get_device_info(self):
+        self.logger.debug(f"ODL({self}): device_info command is received!")
         self.__ensure_port_open()
         cmd = "V1"
         response = self.serial_command(cmd)
@@ -99,6 +138,7 @@ class OdlOzOptics(OpticalDelayLine):
         return device_name, hwd_version
 
     def get_mfg_date(self):
+        self.logger.debug(f"ODL({self}): get manufacturing date command (get_mfg_dates) is received!")
         self.__ensure_port_open()
         cmd = "d?"
         response = self.serial_command(cmd)
@@ -106,12 +146,15 @@ class OdlOzOptics(OpticalDelayLine):
         return date
 
     def echo(self, on_off):
+        self.logger.debug(f"ODL({self}): echo({on_off}) command is received!")
+
         self.__ensure_port_open()
         cmd = "e" + str(on_off)
         response = self.serial_command(cmd)
         return response
 
     def reset(self):
+        self.logger.debug(f"ODL({self}): reset command is received!")
         self.__ensure_port_open()
         cmd = "RESET"
         response = self.serial_command(cmd)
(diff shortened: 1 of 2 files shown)


auspicacious:
@hhatpsi Please assign yourself to https://github.com/moonshot-nagayama-pj/PnPQ/issues/34 and https://github.com/moonshot-nagayama-pj/PnPQ/issues/35 and change the status to "In Review."

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
+    OdlHomeNotCompleted,
+    OdlGetPosNotCompleted,
+    OdlMoveOutofRangeError,
+)
+
 
 
 class OdlOzOptics(OpticalDelayLine):
+    max_move: int
+    """maximum length for moving odl from 0-position, the total path length will be 2x(max_move)"""
```
auspicacious, line 20: As far as I know, docstrings should not be used to document individual variables.

Please see PEP 257 for details on how to use docstrings. https://peps.python.org/pep-0257/

These should be ordinary comments. They should be on the line above the variable and start with `#`. They should not use `"""`.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
         self.conn.baudrate = 9600
         """Basic Communication BaudRate"""
         self.resolution = 32768 / 5.08
         """32768 steps per motor revolution(5.08 mm = 2xDistance Travel or mirror travel per pitch 0.1 inch)"""
-
+        self.logger = logging.getLogger(f"{self}")
         self.command_terminate = "\r\n"
+        self.max_move = 50
+        self.min_move = 0
+        self.logger = logging.getLogger(f"{self}")
```
auspicacious, line 51: You have initialized `self.logger` twice, both on line 47 and on line 51. One should be deleted.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
-    def connect(self):
-        if self.conn.is_open == 0:
+    def __ensure_port_open(self) -> None:
+        if not self.conn.is_open:
+            self.logger.error("disconnected")
+            raise DeviceDisconnectedError(f"{self} is disconnected")
+
+    def reconnect(self):
+        if self.conn.is_open:
+            self.logger.warning(f"Serial connectio to ODL({self}) is already open!")
```
auspicacious, line 66: This should be `info`, not `warning`. There is nothing to warn about here.

In general, `warning` is not a very useful logging level. Most things should be `debug`, `info`, or `error`.

Spelling needs correction.

When logging, do not use f-strings. It is more efficient to use the logger's built in string formatting support. The f-string is computed even if we are not logging `info`-level logs. On the other hand, if we use the built-in string formatting support, the formatted string will not be computed when `info`-level logging is disabled.

```suggestion
            self.logger.warning("Serial connection to ODL(%s) is already open.", self)
```

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
+        else:
+            self.logger.info(f"Openning connection to ODL({self}) device!")
```
auspicacious, line 68: Spelling needs correction: `Openning` -> `Opening`.

See earlier comment about not using f-strings in logging.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
             try:
                 self.conn.open()
-            except Exception as err:
-                raise RuntimeError("Connection failed: " + str(err))
+            except:
+                raise DeviceDisconnectedError(f"{self} is disconnected")
```
auspicacious, line 72: This error is not an accurate description of what is happening.

The old code was more accurate, but it is probably still not correct.

The best thing to do here might be to remove the try/except logic completely. Just let any exceptions from `self.conn.open()` continue up the stack.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
 
     def move(self, dist: float):
-        if not self.conn.is_open:
-            raise RuntimeError("Moving ODL failed: can not connect to ODL device")
-        if dist > 200 or dist < 0:
-            raise Exception("Invalid Move Parameter")
+        self.logger.debug(f"ODL({self}): move({dist}) command has been received")
```
auspicacious, line 75: "has been received" is confusing. Actually, the command has not been received by the ODL yet. In fact, we haven't even sent it yet.

Writing method entry/exit logs manually is not a good idea. The best thing to do would be to [create a decorator](https://duckduckgo.com/?t=ffab&q=python+logging+decorators&ia=web).

But even if we don't want to use a decorator, there are ways to improve this.

For example, if the ODL should be identified in every log line, we could find a way to use logging context and logging filters to add that information to every log line, instead of manually adding it to each line.

But, for now, instead of implementing anything fancy like decorators and logging contexts, **I think the best way to solve this problem is not to log method entry/exit.** Instead, log the commands in `serial_command`. This is much simpler, and gives you almost as much information.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
+        self.__ensure_port_open()
+
+        if dist > self.max_move or dist < self.min_move:
+            raise OdlMoveOutofRangeError(f"ODL({self}): Invalid Move Parameter:{dist}")
         else:
-            self.set_step(int(dist * self.resolution))
+            cmd = "S" + str(int(dist * self.resolution))
```
auspicacious, line 81: This should be an f-string.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
+            response = self.serial_command(cmd)
+
+            #response = self.set_step(int(dist * self.resolution))
```
auspicacious, line 84: Unused code should be deleted, not commented out.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
-            raise Exception("Invalid Move Parameter")
+        self.logger.debug(f"ODL({self}): move({dist}) command has been received")
+        self.__ensure_port_open()
+
+        if dist > self.max_move or dist < self.min_move:
+            raise OdlMoveOutofRangeError(f"ODL({self}): Invalid Move Parameter:{dist}")
         else:
-            self.set_step(int(dist * self.resolution))
+            cmd = "S" + str(int(dist * self.resolution))
+            response = self.serial_command(cmd)
```
auspicacious, line 82: Avoid unnecessary indirection. Don't make a `cmd` variable and then only use it once. Instead, just compose the command directly in this line:
```suggestion
            response = self.serial_command(f"S{whatever}")
```

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
+    """minimum specified length for moving: usually zero"""
+
+    resolution: int
+    """represents minimum specified length for moving on the stage: usually zero"""
+
+    home_timeout: int
+    """required time for completing home in seconds"""
+
+    move_timeout: int
+    """required time for completing home in seconds"""
```
zigen, line 32: home -> move

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
         return response
 
-    def current_step():
-        pass
+    def get_step(self):
+        self.logger.debug(f"ODL({self}): get_step has been started")
+        self.__ensure_port_open()
+        cmd = "S?"
+        response = self.serial_command(cmd)
+        if (response.find("UNKNOWN") >= 0):
```
auspicacious, line 103: The Pythonic way to check if one string is contained in another string is this:

```suggestion
        if 'UNKNOWN' in response:
```

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
 
 class OdlOzOptics(OpticalDelayLine):
+    max_move: int
+    """maximum length for moving odl from 0-position, the total path length will be 2x(max_move)"""
+
+    min_move: int
+    """minimum specified length for moving: usually zero"""
+
+    resolution: int
+    """represents minimum specified length for moving on the stage: usually zero"""
```
zigen, line 26: what's the unit for this? in mm? in um?

src/pnpq/devices/odl_ozoptics_650ml.py
zigen, line 51: why do you need to initialize `self.logger` twice? you should remove this line.

src/pnpq/devices/odl_ozoptics_650ml.py
zigen: and obviously it's not `int`. should be `float`

src/pnpq/devices/odl_ozoptics_650ml.py
zigen, line 66: use the early-return pattern.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
+    def reconnect(self):
+        if self.conn.is_open:
+            self.logger.warning(f"Serial connectio to ODL({self}) is already open!")
+        else:
+            self.logger.info(f"Openning connection to ODL({self}) device!")
             try:
                 self.conn.open()
-            except Exception as err:
-                raise RuntimeError("Connection failed: " + str(err))
+            except:
```
zigen, line 71: why did you remove `Exception as err:` ? it might be helpful for debugging.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -4,63 +4,135 @@
...
-        if not self.conn.is_open:
-            raise RuntimeError("Moving ODL failed: can not connect to ODL device")
-        if dist > 200 or dist < 0:
-            raise Exception("Invalid Move Parameter")
+        self.logger.debug(f"ODL({self}): move({dist}) command has been received")
+        self.__ensure_port_open()
+
+        if dist > self.max_move or dist < self.min_move:
+            raise OdlMoveOutofRangeError(f"ODL({self}): Invalid Move Parameter:{dist}")
         else:
```
zigen, line 80: you don't need this else-clause. because `raise` stops the execution immediately.
just remove `else` and un-indent the following part.

src/pnpq/devices/odl_ozoptics_650ml.py
zigen, line 84: remove useless comment.

src/pnpq/devices/odl_ozoptics_650ml.py
hhatpsi: Because I replace it with DeviceDisconnectedError exception! If we need the previous exception I'll do it!

auspicacious (commented):
In the future, please try to fix one bug at a time for easier review.

This is not a complete review. I only looked at some superficial issues, and I haven't started reviewing the logic yet.

Also, FYI. https://github.com/moonshot-nagayama-pj/PnPQ/issues/27

src/pnpq/devices/odl_ozoptics_650ml.py
hhatpsi: Actually this is the indication of receiving a command (such as move or home) from the host controller point of view. The we translate the command to ODL-device sequence command for example "S+{something}" and "S?" 
 or "GR". 
In this step I'll log the sequence commands in ` serial_command `. For future I'll create the decorator!

auspicacious: I don't know what you mean by "host controller."

This log line marks entry into a function.

If "host controller" means something like `testbed-node`, that code is not part of the PnPQ library. Describing testbed-node actions inside of PnPQ breaks [encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)).

auspicacious:
I'm closing this pull request. We're expecting a smaller pull request that fixes #34 and #35 , and then we can discuss more refactoring after that.

PR #38, opened 2024-07-29
Fix bugs #34 #35 for unifying and error handling of get_step in OzOptics ODL 

hhatpsi:
The bugs #34 and #35 are related. `current_step` is intended to perform the same functionality as `get_step`. Therefore, these two methods have been unified under `get_step`. Basic error handling for the UNKNOWN state has been implemented.

### DIFF
--- a/src/pnpq/devices/odl_ozoptics_650ml.py
+++ b/src/pnpq/devices/odl_ozoptics_650ml.py
@@ -4,6 +4,7 @@
 import serial
 from serial import Serial
 from pnpq.devices.optical_delay_line import OpticalDelayLine
+from pnpq.errors import OdlGetPosNotCompleted
 import time
 
 
--- a/src/pnpq/devices/odl_ozoptics_650ml.py
+++ b/src/pnpq/devices/odl_ozoptics_650ml.py
@@ -48,7 +48,7 @@ class OdlOzOptics(OpticalDelayLine):
         response = self.serial_command(cmd)
         return response
 
-    def get_step(self):
+    def get_step(self) -> int:
         response = self.serial_command(f"S?")
         if "UNKNOWN" in response:
             raise OdlGetPosNotCompleted(
--- a/src/pnpq/devices/odl_ozoptics_650ml.py
+++ b/src/pnpq/devices/odl_ozoptics_650ml.py
@@ -116,12 +116,6 @@ class OdlOzOptics(OpticalDelayLine):
         response = self.serial_command(cmd)
         return response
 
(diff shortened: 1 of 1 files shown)


src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -47,8 +48,14 @@ def set_step(self, value):
         response = self.serial_command(cmd)
         return response
 
-    def current_step():
-        pass
+    def get_step(self):
```
zigen, line 51: add return type hint `def get_step(self) -> int:`

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -47,8 +48,14 @@ def set_step(self, value):
         response = self.serial_command(cmd)
         return response
 
-    def current_step():
-        pass
+    def get_step(self) -> int:
+        response = self.serial_command(f"S?")
```
auspicacious, line 52: The "f-" in "f-string" means "format."

There is nothing to format in this string, therefore, it can just be a normal string.
```suggestion
        response = self.serial_command("S?")
```

src/pnpq/devices/odl_ozoptics_650ml.py
hhatpsi: Done!

src/pnpq/devices/odl_ozoptics_650ml.py
hhatpsi: Done!

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -47,8 +48,14 @@ def set_step(self, value):
...
 
-    def current_step():
-        pass
+    def get_step(self) -> int:
+        response = self.serial_command("S?")
+        if "UNKNOWN" in response:
+            raise OdlGetPosNotCompleted(
+                f"Unknown position for ODL({self}): run find_home() first and then change or get the position"
+            )
+        step = response.split("Done")[0].split(":")[1]
```
auspicacious, line 57: This is existing functionality, and I don't want to change it in this pull request.

However, this is a very strange way of parsing text, especially multi-line text. It's especially strange now that I can see what the text looks like in https://github.com/moonshot-nagayama-pj/PnPQ/issues/39

There is probably a better way to write this that is easier to understand. But we will have to examine that later.

auspicacious (approved):

PR #41, opened 2024-07-29
odl_stub only for forward and backward

TonyT0428:
I paste all the code from [the Thorlabs ODL](https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/devices/odl_thorlabs_kbd101.py), but only modified "init", "step_fforward", "step_backward"


auspicacious:
Please do not copy/paste and commit all of the Thorlabs ODL code here.

**Only** implement the `step_forward` and `step_backward` methods, **as well as tests** for those two methods.

Thank you for adding type hints.

PR #46, opened 2024-07-31
OzOptics ODL-650-ml device response check!

### LINKED ISSUE #45: OzOptics ODL-360 response check
OzOptics ODL-650-m might not provide the complete response during the command execution. This is also related to issue #39, but might not close the bug. 
The maximum calculated delay is related to find_home() command which takes ~3.92 sec. But the delay for moving from position 0mm to position 49mm (max) is ~1.6 seconds. Here we consider max 100 retries for reading response from serial interface with delay of 50 millisecond per retry (total 5 seconds).

hhatpsi:
In OzOptics ODL 650-ml, to verify that the commands are executed accurately, first we should ensure that the response are completely retrieved from the device. This also close #45!


src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -164,11 +165,19 @@ def serial_read(self, retries=10):
...
+        if not got_OK:
+            # means command did not complete before soft timeout!
+            raise RuntimeError(
+                "reading from the device did not complete (soft timeout occured!"
+            )
+
+        self.logger.debug("Device response: %s", device_output)
         return device_output
 
     def serial_command(self, serial_cmd, retries=5):
```
auspicacious, line 178: The `retries` parameter is now unused, I assume it should be removed.
```suggestion
    def serial_command(self, serial_cmd):
```

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -164,11 +165,19 @@ def serial_read(self, retries=10):
                 got_OK = True
             time.sleep(0.05)
             retries -= 1
+
+        if not got_OK:
+            # means command did not complete before soft timeout!
```
auspicacious, line 170: What is a "soft timeout?"

Also, the logic on line 161 is problematic.

```
while self.conn.in_waiting > 0 or (got_OK is False and retries > 0):
```

Let's think about this.

If `in_waiting > 0`, then we can go through all 100 retries, and we can continue to retry 101, because `in_waiting > 0`. The first half of the `or` condition is still true.

As long as there is something `in_waiting`, the loop will continue.

Even if `got_OK == True`, as long as there is something `in_waiting`, the loop will continue.

This doesn't sound like the desired behavior.

I think that the `in_waiting` test should be removed from the if-condition.

Also, [don't use `is` to compare booleans, use `==`](https://stackoverflow.com/questions/4591125/is-it-safe-to-replace-with-is-to-compare-boolean-values). `is` is dangerous. Actually, here, you don't need either one.

```
while (not got_OK) and retries > 0:
```

Line 162 is also worrisome. What happens if `in_waiting` is 0?

auspicacious (commented):
I just took a quick look at the [pySerial documentation](https://pyserial.readthedocs.io/en/latest/pyserial.html) and it says that it features a file-like API.

We should be using the file-like API and the serial library's built-in timeout support.

We should not be trying to loop and read bytes like this. This approach is error-prone and unnecessary.

Specifically, we should use [the `read_until`](https://pyserial.readthedocs.io/en/latest/pyserial_api.html) method to read until we see `Done`.

The `read_until` method can replace this entire function.

auspicacious:
I should add to the above that when using `read_until`, the timeout behavior will probably be different from expected. There's a few details in the pySerial documentation. I don't think that it will be a problem, though.

hhatpsi (commented):
I've attempted to resolve the previous comments!

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -20,9 +21,10 @@ def __init__(
         """Basic Communication BaudRate"""
         self.resolution = 32768 / 5.08
         """32768 steps per motor revolution(5.08 mm = 2xDistance Travel or mirror travel per pitch 0.1 inch)"""
+        self.conn.timeout = 4
```
auspicacious, line 24: Timeouts should be generous... 4 seconds seems quite short.

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -151,24 +153,23 @@ def serial_send(self, serial_cmd):
...
-        while self.conn.in_waiting > 0 or (got_OK is False and retries > 0):
-            device_output += (self.conn.read(self.conn.in_waiting)).decode("iso-8859-1")
-            # command output is complete.
-            if device_output.find("Done") >= 0:
-                got_OK = True
-            time.sleep(0.05)
-            retries -= 1
+        device_output += (self.conn.read_until(expected=bytearray(b"Done"))).decode(
+            "iso-8859-1"
+        )
```
auspicacious, line 163: You can delete line 160 (`device_output = ""`).

```suggestion
        device_output = self.conn.read_until(expected=bytearray(b"Done")).decode("iso-8859-1")
```

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -151,24 +153,23 @@ def serial_send(self, serial_cmd):
...
+        if "Done" in device_output:
+            self.logger.debug("Device read successful: %s", device_output)
+        else:
+            raise RuntimeError("Reading from the device failed (Timeout)!")
```
auspicacious, line 167: This if/else statement confuses two orthogonal actions:

1. Debug logging all captured device output
2. Validating that the read completed successfully.

It's better to just run the debug log statement all of the time.

And we might want to see partial outputs even if debug logging is turned off, so adding it to the RuntimeError might be a good idea.

```suggestion
        self.logger.debug("Device read output: %s", device_output)
        if "Done" not in device_output:
            raise RuntimeError(f"Device read did not complete before timeout. Read output was: {device_output}")
```

src/pnpq/devices/odl_ozoptics_650ml.py
```diff
@@ -151,24 +153,23 @@ def serial_send(self, serial_cmd):
         self.conn.flushOutput()  # flush output buffer, aborting current output and discard all that is in buffer
         self.conn.write(serial_cmd.encode())
 
-    def serial_read(self, retries=10):
+    def serial_read(self):
         # The Python serial "in_waiting" property is the count of bytes available
         # for reading at the serial port.  If the value is greater than zero
         # then we know we have content available.
```
auspicacious, line 159: Delete these comments, they are no longer relevant.

```suggestion
```

src/pnpq/devices/odl_ozoptics_650ml.py
hhatpsi: Done!

src/pnpq/devices/odl_ozoptics_650ml.py
hhatpsi: Done!

src/pnpq/devices/odl_ozoptics_650ml.py
hhatpsi: Done!

auspicacious:
@hhatpsi I can't remember why we didn't finish this PR, let's check in about it.

auspicacious (approved):

Issue #31, opened 2024-07-18
Standardized validation of pull requests for PnPQ


auspicacious:
See the parent epic for details and acceptance criteria: https://github.com/moonshot-nagayama-pj/testbed-meta/issues/13

Most likely, this will be the first PR validation that we implement, which means it will help set the pattern for other repositories. After completing this work, we should update the epic with what we have learned.

TomShen1234:
We should also fix the inconsistencies within the VSCode configurations as part of this.

auspicacious:
Completed fixes in https://github.com/moonshot-nagayama-pj/PnPQ/pull/48

PR #47, opened 2024-08-07
Standardized Check Scripts and Workers for PnPQ

TomShen1234:
Implementation for #31 

The check scripts pretty much works, but many parts of the Github worker and the install dependency scripts are pretty much experimental, and maybe will see some feedbacks regarding their structures. But everything should work at this point (and is tested in a test environment). 

Small known issues and remarks:

The `rye` version check has to be disabled temporarily because it fails for some reason in the Github worker, but actually running `rye sync` do work. I'll leave this to be fixed after the first round of feedback. 

I also replaced the `curl` to `wget` in the `install-rye` script because many systems I've been testing come with archaic version of `curl` which doesn't support the `--output-dir` functionality. 

Finally, instead of installing `shellcheck` and `shfmt` via `apt-get`, it's downloading the current latest version directly off Github release due to the same issues described above, where some systems ships archaic version that has missing features.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,30 @@
+# Automatically run semantic checks, linters, and unit tests
+
+name: Code Checks
+
+on:
+  # pull_requests:
+  workflow_dispatch:
+
+jobs:
+  check-codes:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Install other dependencies
+      run: |
+        chmod +x ./bin/install-dependencies.bash
+        ./bin/install-dependencies.bash
+
+    - name: Install rye
+      run: |
+        chmod +x ./bin/install-rye.bash
+        ./bin/install-rye.bash "/tmp/rye"
+
+    - name: Run check script
+      run: |
+        chmod +x ./bin/check.bash
+        ./bin/check.bash
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   check-codes:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
 
     steps:
     - name: Checkout repository
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   check-codes:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
     - name: Checkout repository
@@ -16,15 +16,12 @@ jobs:
 
     - name: Install other dependencies
       run: |
-        chmod +x ./bin/install-dependencies.bash
         ./bin/install-dependencies.bash
 
     - name: Install rye
       run: |
-        chmod +x ./bin/install-rye.bash
         ./bin/install-rye.bash "$(mktemp -d 'rye-install.XXXXXXXX')"
 
     - name: Run check script
       run: |
-        chmod +x ./bin/check.bash
         ./bin/check.bash || true
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -3,8 +3,8 @@
 name: Code Checks
 
 on:
-  # pull_requests:
-  workflow_dispatch:
+  pull_request:
+    types: [opened, reopened, synchronize]
 
 jobs:
   check-codes:
@@ -22,9 +22,9 @@ jobs:
     - name: Install rye
       run: |
         chmod +x ./bin/install-rye.bash
-        ./bin/install-rye.bash "/tmp/rye"
+        ./bin/install-rye.bash "$(mktemp -d 'rye-install.XXXXXXXX')"
 
     - name: Run check script
       run: |
         chmod +x ./bin/check.bash
-        ./bin/check.bash
+        ./bin/check.bash || true
--- a/README.md
+++ b/README.md
@@ -4,7 +4,9 @@ PnPQ is a python library package for controlling optical devices in quantum netw
 
 ## Prerequisites
 
-Install (rye)[https://rye.astral.sh/].
+- (rye)[https://rye.astral.sh/].
+- shellcheck
+- shfmt
 
 ## Test and Debug
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,23 @@
 # pnpq
-Plug and Play with Quantum! 
+Plug and Play with Quantum!
 PnPQ is a python library package for controlling optical devices in quantum networks
 
+## Requirement
+
+Install (rye)[https://rye.astral.sh/].
+
+## Test and Debug
+
+Since PnPQ is a device driver library, there is no main interface available to run.
+
+Instead, unit tests are available, and can be executed with: `rye test`.
+
+## Making Contributions
+
+Before you commit, ensure that `check.sh` does not output any error. This script checks for formatting errors as well as semantics in Python and shell scripts.
+
+### Formatting
+
+For Python, Black will be used as the standard formatter for this repository. It is part of the recommended plugins, and configured as the default formatter in VSCode.
+
+In VSCode, simply run `>Format Document` in the top menu bar with the Python file opened.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 Plug and Play with Quantum!
 PnPQ is a python library package for controlling optical devices in quantum networks
 
-## Requirement
+## Prerequisites
 
 Install (rye)[https://rye.astral.sh/].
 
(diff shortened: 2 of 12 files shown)


bin/install-rye.bash
```diff
@@ -0,0 +1,84 @@
...
+# Add rye into PATH
+source "${HOME}/.rye/env"
+
+# Add rye to PATH for Github Actions (persist path)
+# shellcheck disable=SC2310
+if is_set GITHUB_ENV; then
+  # Write the new path to Github env
+  # shellcheck disable=SC2154
+  stdmsg "PATH=${PATH}" >>"${GITHUB_ENV}"
+fi
```
auspicacious, line 84: I guess this will work.

But you should definitely move the `source` inside the `GITHUB_ENV` check, otherwise it is just going to sow confusion. The `source` is essentially a no-op when run outside of GitHub Actions, since any changes to the environment caused by `source` will be lost as soon as this script exits.

```suggestion
# Add rye to PATH for Github Actions (persist path)
# shellcheck disable=SC2310
if is_set GITHUB_ENV; then
  source "${HOME}/.rye/env"
  # Write the new path to Github env
  # shellcheck disable=SC2154
  stdmsg "PATH=${PATH}" >>"${GITHUB_ENV}"
fi
```

bin/install-dependencies.bash
```diff
@@ -0,0 +1,54 @@
...
+
+# Install wget and shellcheck
+apt-get update
+apt-get install wget xz-utils -y
+
+# Install shellcheck
+wget https://github.com/koalaman/shellcheck/releases/download/v0.10.0/shellcheck-v0.10.0.linux.aarch64.tar.xz --output-document /tmp/shellcheck.tar.xz
+tar xJf /tmp/shellcheck.tar.xz --directory /tmp
+mv /tmp/shellcheck-v0.10.0/shellcheck /usr/local/bin/shellcheck
+rm -rf /tmp/shellcheck-v0.10.0 /tmp/shellcheck.tar.xz
```
auspicacious, line 50: I explained in another comment why writing files directly into the global `/tmp` is a bad idea.

This is a good place to use the exit trap. Essentially, make a temporary directory when the script starts running and then `rm -rf` the directory in the trap when execution is finished.

bin/install-dependencies.bash
```diff
@@ -0,0 +1,54 @@
...
+# Update package list
+apt-get update
+
+# Install wget and shellcheck
+apt-get update
+apt-get install wget xz-utils -y
+
+# Install shellcheck
+wget https://github.com/koalaman/shellcheck/releases/download/v0.10.0/shellcheck-v0.10.0.linux.aarch64.tar.xz --output-document /tmp/shellcheck.tar.xz
+tar xJf /tmp/shellcheck.tar.xz --directory /tmp
```
auspicacious, line 48: Good use of the `--directory` option! But, note well what the manpage says:

```
       -C, --directory=DIR
              Change to DIR before performing any operations.  **This option is order-sensitive**, i.e. it affects all options that follow.
```

Also, [the version of `tar` shipped with MacOS](https://ss64.com/mac/tar.html) (which is based on BSD tar, not GNU tar) only knows about `-C`, not `--directory`, so it's often safer to use `-C` here.

So, in order to make sure that all tar commands use the right working directory, and the script has a fighting chance of working cross-platform, write it like this:

```
tar -C "${temp_download_dir}" -xJf shellcheck.tar.xz
```

README.md
```diff
@@ -1,4 +1,23 @@
 # pnpq
-Plug and Play with Quantum! 
+Plug and Play with Quantum!
 PnPQ is a python library package for controlling optical devices in quantum networks
 
+## Requirement
+
+Install (rye)[https://rye.astral.sh/].
```
auspicacious, line 7: Missing a couple of prerequisites here.

.github/workflows/check.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    - name: Install rye
+      run: |
+        chmod +x ./bin/install-rye.bash
+        ./bin/install-rye.bash "/tmp/rye"
+
+    - name: Run check script
+      run: |
+        chmod +x ./bin/check.bash
+        ./bin/check.bash
```
auspicacious, line 30: Until the code actually passes all of the lint checks, we can put this as 
```suggestion
        ./bin/check.bash || true
```

so that it always passes.

TomShen1234:
I knew many parts were somewhat hacky, so these feedbacks are very appreciated. I'll implement the changes

bin/check.bash
```diff
@@ -0,0 +1,69 @@
...
+# version_string=$(rye --version | head -n 1 | cut -d ' ' -f 2)
+# stdmsg "Rye version: ${version_string}"
+
+# Sync with rye
+stdmsg "Running rye sync..."
+rye sync
+
+# Activate virtual environment
+stdmsg "Activating virtual environment..."
+source .venv/bin/activate
```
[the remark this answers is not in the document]
TomShen1234: I guess we could deactivate it at the end to restore the `$PATH`, but it's really not necessary here as everything afterward depends on the virtual environment.

bin/check.bash
```diff
@@ -0,0 +1,69 @@
...
+
+# Run black check
+stdmsg "Checking Python code formatting with black..."
+black --check --diff src tests
+
+# Shell check
+# Recursively loop through all files and find all files with .sh extension and run shellcheck
+stdmsg "Checking Shell scripts with shellcheck..."
+find . -type f \( -name "*.sh" -o -name "*.bash" \) -print0 | xargs -0 shellcheck --source-path .:"${HOME}" --enable=all --external-sources
```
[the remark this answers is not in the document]
TomShen1234: It's for the `source '${HOME}/.rye/env'` line. shellcheck cannot figure out what `$HOME` is for that line, therefore the `source-path` option is used to tell it to look in the home folder for that script. But I guess I could also just deactivate that particular check for the source line, since there's only 1 of them.

bin/install-dependencies.bash
```diff
@@ -0,0 +1,54 @@
...
+is_set() {
+  [[ -n ${!1+x} ]]
+}
+
+# Update package list
+apt-get update
+
+# Install wget and shellcheck
+apt-get update
+apt-get install wget xz-utils -y
```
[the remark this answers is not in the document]
TomShen1234: Let me try updating my test environment to a newer version (or should I say, 'less archaic' version 😅), I think it may be due to using an outdated version of Ubuntu for testing resulted in an old version of curl being installed.

bin/install-dependencies.bash
TomShen1234: I'll try to install these via `apt-get` again.

bin/install-dependencies.bash
```diff
@@ -0,0 +1,54 @@
...
+
+# Install shellcheck
+wget https://github.com/koalaman/shellcheck/releases/download/v0.10.0/shellcheck-v0.10.0.linux.aarch64.tar.xz --output-document /tmp/shellcheck.tar.xz
+tar xJf /tmp/shellcheck.tar.xz --directory /tmp
+mv /tmp/shellcheck-v0.10.0/shellcheck /usr/local/bin/shellcheck
+rm -rf /tmp/shellcheck-v0.10.0 /tmp/shellcheck.tar.xz
+
+# Install shfmt
+wget https://github.com/mvdan/sh/releases/download/v3.8.0/shfmt_v3.8.0_linux_amd64 --output-document /usr/local/bin/shfmt
+chmod +x /usr/local/bin/shfmt
```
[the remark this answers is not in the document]
TomShen1234: I'll try that again with shellcheck, but for shfmt, `apt-get` cannot find the package `golang-mvdan-sh`, unless I'm missing something. I checked

bin/install-dependencies.bash
```diff
@@ -0,0 +1,55 @@
...
+  if [[ ${exit_status} -ne 0 ]]; then
+    errmsg 'There is an error installing the dependencies.'
+    exit 1
+  fi
+}
+trap trap_exit EXIT
+
+# Update package list and install curl and shellcheck
+sudo apt-get update
+sudo apt-get install curl shellcheck -y
```
auspicacious, line 35: By the way, I noticed that [the 24.04 runner image comes with curl and shellcheck preinstalled](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md).

It won't hurt performance to leave them in this `apt-get` command. `apt-get` doesn't reinstall already-installed packages unless you pass the `--reinstall` flag.

auspicacious (commented):
The only issue that needs resolving before approval is to move to the `24.04` image and use `apt-get` to install `shfmt`.

TomShen1234:
Hmmm, just did that, didn't behave as expected. Still cannot find the package.

TomShen1234:
Looks like `sudo apt-get install shfmt -y` is the way to go on Ubuntu 24.04.

auspicacious (approved):
I missed the fact that the package is `golang-mvdan-sh-dev` on Ubuntu while it's `golang-mvdan-sh` on Debian. That should have been a sign it wouldn't work properly.

`-dev` packages provide whatever a developer would need to build code that requires a library. For C projects, this is mostly header files, and then the code you build would link against the prebuilt shared library provided by the corresponding non- `-dev` package. I'm not sure how this works for Go, where shared libraries are not the norm.

Anyway, thank you, I will merge this and look at fixing the reported errors soon. There's some code comments I think could be improved on this PR, but we can discuss them some other time.

PR #50, opened 2024-08-14
Clean up package init

auspicacious:
This code may require refactoring of any other code using PnPQ, depending on how it was `import`ing PnPQ classes. It is likely that more refactoring will eventually follow these changes.

The `__init__` files were being used to expose some of the device implementation classes as members of `pnpq`, and, confusingly, sometimes, also `pnpq.devices`.

By removing the `import` statements and `__all__` variables from `__init__.py`, we ensure that by default nothing will be imported into any module that calls `import pnpq` or `from pnpq import *`.

Instead, we should simply use fully-qualified imports, such as `from pnpq.devices.waveplate_stub import WaveplateStub`.

I think that this is in line with [the advice in the Python tutorial](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package), and it makes the code structure much easier to understand.

I also found and fixed a few other things along the way.

Full list of changes:

* `pnpq.utils` has been deleted. It duplicated the functionality in `pnpq.devices.utils` and appears to be unused.
* Unit tests have been renamed to begin with `test_` rather than end with `_test`. This follows common pytest form.
* Very basic unit tests have been added for all device modules and the utils module. These tests only validate that when no device is physically connected, an exception is thrown.
* The `__version__` variable now works. See changes in `pyproject.toml`.
* An issue in `optical_delay_line` where a non-existent `serial` module was being referenced is fixed.

Tested by running unit tests and the `example.py` script.


zigen (approved):

TomShen1234 (approved):

tests/test_waveplate_thorlabs_kb10crm.py
```diff
@@ -0,0 +1,9 @@
+from pnpq.devices.waveplate_thorlabs_kb10crm import Waveplate
+from pnpq.errors import DevicePortNotFoundError
+
+import pytest
+
+
+def test_disconnected_initialization():
+    with pytest.raises(DevicePortNotFoundError):
+        Waveplate("ABC", "DEF")
```
hhatpsi, line 9: For Thorlabs K10CR1 waveplate rotator, the device serial number starts with '55' and is in this format: 55xxxxxx

tests/test_odl_thorlabs_kbd101.py
```diff
@@ -0,0 +1,8 @@
+from pnpq.devices.odl_thorlabs_kbd101 import OdlThorlabs
+
+import pytest
+
+
+def test_disconnected_initialization():
+    with pytest.raises(RuntimeError):
+        OdlThorlabs("ABC", "DEF")
```
hhatpsi, line 8: Based on type of device in Thorlabs, the serial number starts with a specific number. It might be useful to check it in the unit test. For example for KB101, the serial number starts with '28' and with this format: '28xxxxxx'

src/pnpq/utils.py
```diff
@@ -1,15 +0,0 @@
-from serial.tools.list_ports import comports as list_comports
-import logging
-
-logger = logging.getLogger("utils")
-
-
-def get_available_port(device_serial_number: str) -> str | None:
-    logger.debug(f"get_available_port(serial_number: {device_serial_number})")
```
hhatpsi, line 8: We have almost used get_available_port() for connecting to a specific device regardless of how many devices are connected to the host controller. After removing this file 'utils.py', I could not find the alternative of get_available_port().

tests/test_devices_utils.py
```diff
@@ -0,0 +1,9 @@
+import pnpq.devices.utils as utils
+
+
+def test_get_available_port_no_available_ports():
+    assert utils.get_available_port("ABC") is None
+
```
hhatpsi, line 6: Where is get_available_port() defined?

auspicacious: See line 1 and see the PR summary.

tests/test_odl_thorlabs_kbd101.py
auspicacious: These values are intentionally invalid. I am testing failure to find a device.

The serial number format is useful information and that's good to know, but it's not relevant to us right now. In the future, we could parse the serial number and throw an exception if it doesn't start with a known value like `28`.

tests/test_waveplate_thorlabs_kb10crm.py
auspicacious: That's good to know for the future.

src/pnpq/utils.py
auspicacious: Please see the PR summary.

> pnpq.utils has been deleted. It duplicated the functionality in pnpq.devices.utils and appears to be unused.

auspicacious:
@hhatpsi I'm going to merge this, but we can continue the discussion about utils if necessary.

PR #52, opened 2024-09-02
Add mypy to checks, correct static typing errors

auspicacious:
Adding [mypy](https://mypy.readthedocs.io/en/stable/) was very useful for finding errors and misunderstandings.

I did have to change the behavior of code in a few places, and some of the type annotations I added in code we don't really use were guesses. However, I think that everything should be functionally equivalent, _except_ that I removed the `config_file` parameter from a couple of device constructors because it wasn't being used.


TomShen1234 (approved):
I checked the stub devices and the tests. These parts look fine to me.

auspicacious:
I checked testbed-node to make sure that `config_file` wasn't being used, and it seems to be OK.

The only thing I'm not really sure about is some of the functions that we don't use in `odl_ozoptics_650ml.py` where I've set `parameter` to be `int`. That was a guess, and I haven't been able to find the corresponding commands in the manual. Most likely these will be refactored away in the future, and I'm not too worried about it.

hhatpsi (approved):

PR #54, opened 2024-09-18
Modify the required check for custom_rotate

### LINKED ISSUE #53: custom_rotate failed with calibration_offset = 0
When there is a default configuration for waveplate (i.e `calibration_offset = 0`), after homing `relative_home` will be set to zero in `custom_home()`. 
But for 'custom_rotate()' the following check in waveplate_controller.py is not logically correct: 
`    def custom_rotate(self, degree):
        """Rotattion with customized home!"""

        if not self.relative_home:
            self.logger.error(
                "relative_home parameter should be defined befor custom rotation"
            )
            raise WavePlateCustomRotateError(
                f"Waveplate({self}) relative_home can not be completed"
            )
`

hhatpsi:
This change will fix #53

### DIFF
--- a/src/pnpq/devices/waveplate_thorlabs_kb10crm.py
+++ b/src/pnpq/devices/waveplate_thorlabs_kb10crm.py
@@ -343,7 +343,7 @@ class Waveplate:
     def custom_rotate(self, degree: float | int) -> None:
         """Rotation with customized home!"""
 
-        if not self.relative_home:
+        if self.relative_home is None:
             self.logger.error(
                 "relative_home parameter should be defined befor custom rotation"
             )
--- a/src/pnpq/devices/waveplate_thorlabs_kb10crm.py
+++ b/src/pnpq/devices/waveplate_thorlabs_kb10crm.py
@@ -8,7 +8,6 @@ from pnpq.errors import (
     WavePlateMoveNotCompleted,
     WavePlateHomedNotCompleted,
     WavePlateGetPosNotCompleted,
-    WavePlateCustomRotateError,
     WaveplateInvalidStepsError,
     WaveplateInvalidDegreeError,
     WaveplateEnableChannelError,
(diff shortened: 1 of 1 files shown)


src/pnpq/devices/waveplate_thorlabs_kb10crm.py
```diff
@@ -343,7 +343,7 @@ def custom_home(self, degree: float | int) -> None:
     def custom_rotate(self, degree: float | int) -> None:
         """Rotation with customized home!"""
 
-        if not self.relative_home:
+        if self.relative_home is None:
```
auspicacious, line 346: Unfortunately, this isn't going to work. I should have noticed this myself, but I'm still learning the rules of Python variable definition myself, and I forgot about "truthy" and "falsy" values as well.

I think that the best way to explain it is with some simple example code:

```python
#!/usr/bin/env python

class MyClass():
    set_variable = "class_set_variable"
    unset_variable: str

    def __init__(self):
        self.set_variable = "instance_set_variable"

    def uncalled_function(self):
        self.unset_variable = "unset_variable_value"

    def print_values(self):
        print("class-level set variable: " + MyClass.set_variable)
        print("instance-level set variable: " + self.set_variable)

        # print("class-level unset variable: " + MyClass.unset_variable)
        # print("instance-level unset variable" + self.unset_variable)
        # if not self.unset_variable:
        #    print("not unset_variable")
        if self.unset_variable is None:
            print("unset_variable is None")

if __name__ == "__main__":
    my_obj = MyClass()
    my_obj.print_values()
```

You can play around with the commented-out code a little to see the different errors that you get.

This becomes even more confusing if you change `unset_variable: str` to `unset_variable: str = None` ! That will create a class-level variable called `unset_variable`, and then `self.unset_variable` will work as you expect.

However, this won't pass a type checker. `None` is not type `str`.

And then Python [dataclasses](https://docs.python.org/3/library/dataclasses.html) work according to different rules again. I won't explore that here.

So, what is the solution?

We need to ensure that `self.relative_home` always has a value.

For now, the easiest way to do this is to add a line to `__init__`:

```
self.relative_home = 0.0
```

and then completely remove lines 346 to 353.

hhatpsi: Thank you for providing such a helpful description.

auspicacious (approved):
Thank you, hopefully this will work properly.

Issue #51, opened 2024-09-02
Abstract interface for motorized polarization controllers and MPC320 driver


auspicacious:
We want to support motorized polarization controllers such as the Thorlabs MPC320.

This is a good chance for us to see if we can successfully implement an abstract base class as described in https://github.com/moonshot-nagayama-pj/PnPQ/issues/29 before we attempt to refactor existing code.

## Acceptance criteria
* Third-party libraries https://thorlabs-apt-device.readthedocs.io/en/latest/index.html , https://pypi.org/project/pyThorlabsAPT/ https://github.com/UniNE-CHYN/thorpy and https://pylablib.readthedocs.io/en/latest/devices/Thorlabs_misc.html https://instrumentkit.readthedocs.io/en/latest/ are evaluated for comparison and possible use
* Abstract base class provides an interface that could be used for any similar device, including ones from competing manufacturers
* Implementation for MPC320 is complete
* Manual testing against physical device is complete
* Code passes all static analysis checks and requirements in https://github.com/moonshot-nagayama-pj/PnPQ/issues/29

## Good to have
* Some form of automated testing is possible, maybe by mocking the underlying pySerial library.
* Using a library that allows for easy conversion of scientific units may be desirable, and those should be evaluated.

## References
* https://www.thorlabs.com/thorproduct.cfm?partnumber=MPC320
* https://www.notion.so/Thorlabs-Motorized-Polarization-Controller-9fb30b33750545a9916acdb87216e46e
* https://github.com/moonshot-nagayama-pj/playground/blob/main/amin/devel-testbed-tools/polc_thorlabs_CLI_2024_08_22.py

auspicacious:
Some notes on third-party libraries:

* thorlabs-apt-device is actively maintained and cross-platform. It might be useful for us to use this library in implementation. Unfortunately it does not have type hints yet.
* pyThorlabsAPT is focused on Windows and uses the proprietary Thorlabs DLL, so it is not useful to us.
* thorpy does not appear to be maintained.
* pylablib seems like it might be difficult for us to extend compared to thorlabs-apt-device. Also, it seems to be basing its implementations on the SCPI protocol, not APT, so I'm not sure that it's appropriate for our devices.

auspicacious:
Wow, there are a lot of Python libraries related to scientific units.

## No longer maintained or considered not ready for production

We should avoid using these.

[...]

PR #55, opened 2024-10-04
Add MPC320 support and framework for managing Thorlabs APT devices

auspicacious:
See https://github.com/moonshot-nagayama-pj/PnPQ/issues/51 for additional background. I think that we should consider this as closes https://github.com/moonshot-nagayama-pj/PnPQ/issues/51 , although the requirements have changed during development. (I will note that in the issue).

I've made a couple more changes to the logic since the demo on 10-03. In particular, I noticed that I was not acquiring the `tx_connection_lock` in the appropriate place in the ordered_sender thread.

If you are connected to the hardware, you can run tests by running `pytest hardware_tests`. You will need to adjust the serial number in `hardware_tests/test_polarization_controller_thorlabs_mpc320.py` to match the actual serial number of the device.

I am running Linux; @hhatpsi , will this code work on MacOS?

I have tried to mark some of the major unfinished work with `TODO`. There is probably still more that is missing or could be improved. The code also needs documentation.

My suggestion is that if @TomShen1234 can run the hardware tests against this branch, and there are no major problems found while @hhatpsi and @TomShen1234 review this PR, that we merge this PR and then continue work in additional, smaller PRs.

### DIFF
--- a/bin/check.bash
+++ b/bin/check.bash
@@ -53,7 +53,7 @@ mypy
 
 # Run black check
 stdmsg "Checking Python code formatting with black..."
-black --check --diff src tests
+black --check --diff src tests hardware_tests
 
 # Run shellcheck
 # Recursively loop through all files and find all files with .sh extension and run shellcheck
--- a/hardware_tests/conftest.py
+++ b/hardware_tests/conftest.py
@@ -1,6 +1,10 @@
 import structlog
+import sys
 
 from pathlib import Path
+from pnpq.events import Event
+from types import TracebackType
+from typing import Any
 
 
 def find_project_dir(path: Path) -> Path:
@@ -26,3 +30,17 @@ structlog.configure(
         file=target_dir.joinpath("hardware_tests.log").open("a")
     ),
 )
+
+log = structlog.get_logger()
+
+
+def excepthook(
+    exception_type: type[BaseException],
+    e: BaseException,
+    traceback: TracebackType | None,
+) -> Any:
+    log.error(event=Event.UNCAUGHT_EXCEPTION, exc_info=e)
+    sys.__excepthook__(exception_type, e, traceback)
+
+
+sys.excepthook = excepthook
--- a/hardware_tests/conftest.py
+++ b/hardware_tests/conftest.py
@@ -1,0 +1,28 @@
+import structlog
+
+from pathlib import Path
+
+
+def find_project_dir(path: Path) -> Path:
+    if path.is_dir():
+        for contained_path in path.iterdir():
+            if contained_path.name == "pyproject.toml":
+                return path
+    return find_project_dir(path.parent)
+
+
+target_dir = find_project_dir(Path(__file__).resolve()).joinpath("target")
+target_dir.mkdir(exist_ok=True)
+
+structlog.configure(
+    processors=[
+        structlog.processors.add_log_level,
+        structlog.processors.StackInfoRenderer(),
+        structlog.dev.set_exc_info,
+        structlog.processors.TimeStamper(fmt="iso", utc=True),
+        structlog.processors.JSONRenderer(),
+    ],
+    logger_factory=structlog.WriteLoggerFactory(
+        file=target_dir.joinpath("hardware_tests.log").open("a")
+    ),
+)
--- a/hardware_tests/conftest.py
+++ b/hardware_tests/conftest.py
@@ -1,5 +1,6 @@
 import structlog
 import sys
+import threading
 
 from pathlib import Path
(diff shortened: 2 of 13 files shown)


auspicacious:
@master-of-ppap you should look at this code as well.

TomShen1234:
I tried the codes and the hardware test. 
1. The tests (by extension, the code) don't quite work on macOS. The reason is that the serial device does not have an interface exposed in the `/dev` folder. I presume this is a drivers issue? But I had a quick peek around the internet and cannot find a suitable driver. 
2. I used a Linux virtual machine to test the code. Works fine for me. I'm wondering if it's not possible to get it working directly on macOS, is it possible to setup something like Docker? Using a VM requires a lot of work just to setup the dev environment every time some changes needs to be made.
3. It looks like we need to modify the serial number within the code. I'm thinking if it's better if we expose it as an option to be passed in the hardware test. I think `pytest` supports this.
4. Shall we do something about the `example.py` in the root directory? Seems like that's also a hardware test. If it's actually an example, maybe it's better to go in the readme?

hhatpsi:
> * The tests (by extension, the code) don't quite work on macOS. The reason is that the serial device does not have an interface exposed in the `/dev` folder. I presume this is a drivers issue? But I had a quick peek around the internet and cannot find a suitable driver.

You can find the FTDI driver for macOS from [here](https://ftdichip.com/drivers/vcp-drivers/). You can also find the installation guide document [here](https://www.ftdichip.com/Support/Documents/InstallGuides/Mac_OS_X_Installation_Guide.pdf).  I've tried just one time to install one of the driver from the list for one other device, but It didn't work.  I think we need to apply additional configuration!

auspicacious:
>     1. The tests (by extension, the code) don't quite work on macOS. The reason is that the serial device does not have an interface exposed in the `/dev` folder. I presume this is a drivers issue? But I had a quick peek around the internet and cannot find a suitable driver.

Yeah, I wonder about this. I [did a quick search](https://duckduckgo.com/?t=ffab&q=macos+usb+serial+device&ia=web) and indications are that there should be files that look something like `/dev/tty`. Another search result indicates that FTDI USB to serial chips (which we're using) work even on M1 Mac. I'll see if I can work this out with a Mac user tomorrow.

>     2. I used a Linux virtual machine to test the code. Works fine for me. I'm wondering if it's not possible to get it working directly on macOS, is it possible to setup something like Docker? Using a VM requires a lot of work just to setup the dev environment every time some changes needs to be made.

Yes, if you were able to forward your USB devices into a VM, most likely you can forward them into a Docker host VM, then into a Docker container inside this VM. We can set up VSCode integration like we have for other projects. It is theoretically a little tricky since we don't know the USB device filename we want to forward before we run the program.

FWIW we could also [directly integrate VSCode with a plain VM](https://code.visualstudio.com/docs/remote/ssh), AFAIK. ([emacs can do this too](https://www.gnu.org/software/emacs/manual/html_node/emacs/Remote-Files.html)). In general, I think this is a pretty good option for developers to have a flexible, Linux-based development environment, especially if the developer is using Windows.

This is all assuming we can't figure out why it's not working on MacOS.

Anyway, this is all for another PR.

By the way, another thing we don't support in the current PR is connecting by port name. You can only connect by serial number. We may need to support both in the future.

>     3. It looks like we need to modify the serial number within the code. I'm thinking if it's better if we expose it as an option to be passed in the hardware test. I think `pytest` supports this.

OK, but let's do that in another PR.

>     4. Shall we do something about the `example.py` in the root directory? Seems like that's also a hardware test. If it's actually an example, maybe it's better to go in the readme?

As we move existing Thorlabs devices into this new framework, we will write some test files for them. That's probably when we should delete `example.py`.

TomShen1234:
> Another search result indicates that FTDI USB to serial chips (which we're using) work even on M1 Mac. I'll see if I can work this out with a Mac user tomorrow.

Interestingly enough, it shows up in the Mac's USB device manager, with the serial number also being present and correct. It's just that python's serial library can't find it (presumably because the device file didn't show up). I even did a diff of `/dev` before and after plugging it in, no difference.

> Yes, if you were able to forward your USB devices into a VM, most likely you can forward them into a Docker host VM, then into a Docker container inside this VM. We can set up VSCode integration like we have for other projects. It is theoretically a little tricky since we don't know the USB device filename we want to forward before we run the program.

For today, I installed the SSH plugin in VSCode and managed to get it working efficiently enough once it's setup. 

Since this is not really a Python's problem (probably due to Mac's driver), and we'll fix others in separate PRs, I'll approve this for now.

TomShen1234 (approved):

PR #61, opened 2024-10-15
Add pint to MPC320

### LINKED ISSUE #57: Support Pint unit library for MPC320
We are currently using [the Pint unit library](https://pint.readthedocs.io/en/stable/) in one place, [to represent milliamps in an APT message](https://github.com/moonshot-nagayama-pj/PnPQ/pull/55/files#diff-44611a3a21da3f1f28623db0df50b892df6daa4bc3e912dc760a8cdc825181fbR468).

Let's evaluate if it makes sense for us to use this in a slightly more complex use case: mapping standard angle values like radians or degrees to the motor step units used to control APT devices.

## Acceptance criteria

* Functions that accept an angle in the MPC320 class are modified to accept a Pint quantity instead. Currently known values are move_absolute and the home parameter in set_params.
* This value is correctly translated into the MPC320's units
* This value is logged in a useful, readable way (presumably in degrees?)
* Hardware tests are updated to validate this behavior
* We decide if we want to continue using these units

TomShen1234:
Closes #57 

I didn't test this code on real hardware since I didn't have access to one from writing this software. It would be nice if @auspicacious can try it (or I can try it on Friday). 

I implemented pint for `move_absolute` function (in parameter) as well as the `home_position` of the `PolarizationControllerParams` struct.

### DIFF
--- a/hardware_tests/test_polarization_controller_thorlabs_mpc320.py
+++ b/hardware_tests/test_polarization_controller_thorlabs_mpc320.py
@@ -1,3 +1,4 @@
+from pint import DimensionalityError
 import pytest
 
 from pnpq.apt.protocol import ChanIdent
@@ -73,9 +74,12 @@ def test_move_absolute(device: PolarizationControllerThorlabsMPC320) -> None:
     # device.move_absolute(ChanIdent.CHANNEL_3, 0 * ureg.degree)
 
 
-def test_illegal_angles(device: PolarizationControllerThorlabsMPC320) -> None:
+def test_invalid_angle_inputs(device: PolarizationControllerThorlabsMPC320) -> None:
     device.identify(ChanIdent.CHANNEL_1)
 
     with pytest.raises(ValueError):
         device.move_absolute(ChanIdent.CHANNEL_1, 171 * ureg.degree)
         device.move_absolute(ChanIdent.CHANNEL_1, -1 * ureg.degree)
+
+    with pytest.raises(DimensionalityError):
+        device.move_absolute(ChanIdent.CHANNEL_1, 1 * ureg.meter)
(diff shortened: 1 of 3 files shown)


auspicacious:
@TomShen1234 Don't forget to run the check script locally.

TomShen1234:
The hardware test works

auspicacious (commented):
Apparently [illegal angles are a problem in several sports](https://duckduckgo.com/?t=ffab&q=illegal+angle&ia=web). "Illegal" is one of the most confusing words in computer programs. I would suggest calling the angle "out of bounds" instead.

I'm not sure about converting `DimensionalityError` to `ValueError`. I didn't do this in [the other place we used Pint](https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/apt/protocol.py#L438). Probably during our earlier discussion I either didn't realize that `DimensionalityError` was thrown during the conversions, or I expected that you were going to call `.check` ahead of time.

I think it is probably better to let `DimensionalityError` bubble up. This creates less work for us since we don't have to remember to catch and wrap these everywhere.

If calling `.check` creates a `DimensionalityError` with a more informative error message, maybe call that before trying the conversions? Otherwise, just get rid of the try/catch.

TomShen1234:
`.check` simply returns true/false, without failing. I just removed the try clause, it will automatically throw a `DimensionalityError` if the conversion fails, with the default error message.

auspicacious:
> `.check` simply returns true/false, without failing.

@TomShen1234 Oof, that's my mistake for reading the documentation on the `@check` function decorator and not the documentation on the `.check()` function. (Although that seems like an API design problem to me, too, Pint should really throw an exception in either situation). I'll fix the check in `protocol.py` myself tomorrow to throw `DimensionalityError` somehow.

auspicacious (approved):

Issue #56, opened 2024-10-05
Complete implementation of APT messages for MPC320


auspicacious:
The [APT protocol documentation](https://www.thorlabs.com/Software/Motion%20Control/APT_Communications_Protocol.pdf) contains the full list of messages that can be used with the MPC320 polarization controller.

We have also listed these messages [here in UnimplementedAptMessageId](https://github.com/moonshot-nagayama-pj/PnPQ/pull/55/files#diff-44611a3a21da3f1f28623db0df50b892df6daa4bc3e912dc760a8cdc825181fbR43).

## Acceptance criteria

* All messages implemented in `protocol.py`
* Unit tests written for all messages, following the existing pattern
* Comments added for anything that is unusual or hard to understand from the official APT documentation

auspicacious:
@master-of-ppap After https://github.com/moonshot-nagayama-pj/PnPQ/pull/55 is merged, I think that this is a good place for you to start.

master-of-ppap:
For the first acceptance criteria, should we try to implement the actual functionality of the messages? Or would something simple like the DISCONNECT class be fine for now?
`@dataclass(frozen=True, kw_only=True)
class AptMessage_MGMSG_HW_DISCONNECT(AptMessageHeaderOnlyNoParams):
    message_id = AptMessageId.MGMSG_HW_DISCONNECT`

auspicacious:
@master-of-ppap If I understand your question correctly, the question is whether we need to understand the content of the messages when we write these classes.

The answer is yes. The reason we are implementing these classes is to make communication with the devices more legible.

Also, not all of the messages are simple 6-byte messages.

For example, for `MGMSG_MOT_GET_POSCOUNTER` has an additional 6-byte data packet. We have to build a custom class to understand this message.

My suggestion is to start with a simple message, implement it, and open a PR. Don't try to do all of the message types all at once, and ask questions early if you can't find answers by looking at the Python documentation.

PR #63, opened 2024-10-18
Implement more messages

master-of-ppap:
Partial implementation of #56

Implemented SET_POSCOUNTER, REQ_POSCOUNTER, and GET_POSCOUNTER messages and their respective tests

### DIFF
--- a/src/pnpq/apt/protocol.py
+++ b/src/pnpq/apt/protocol.py
@@ -27,6 +27,10 @@ class AptMessageId(int, Enum):
 
     MGMSG_MOD_IDENTIFY = 0x0223
 
+    MGMSG_MOT_SET_POSCOUNTER = 0x0410
+    MGMSG_MOT_REQ_POSCOUNTER = 0x0411
+    MGMSG_MOT_GET_POSCOUNTER = 0x0412
+
     MGMSG_MOT_ACK_USTATUSUPDATE = 0x0492
     MGMSG_MOT_GET_USTATUSUPDATE = 0x0491
     MGMSG_MOT_REQ_USTATUSUPDATE = 0x0490
@@ -51,10 +55,6 @@ class UnimplementedAptMessageId(int, Enum):
     """
 
     # TODO (see docstring)
-    MGMSG_MOT_GET_POSCOUNTER = 0x0412
-    MGMSG_MOT_REQ_POSCOUNTER = 0x0411
-    MGMSG_MOT_SET_POSCOUNTER = 0x0410
-
     MGMSG_MOT_MOVE_JOG = 0x046A
     MGMSG_MOT_MOVE_STOP = 0x0465
     MGMSG_MOT_MOVE_STOPPED = 0x0466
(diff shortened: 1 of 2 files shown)


src/pnpq/apt/protocol.py
```diff
@@ -688,6 +688,117 @@ class AptMessage_MGMSG_MOD_IDENTIFY(AptMessageHeaderOnlyChanIdent):
...
+            self.destination_serialization,
+            self.source,
+            self.chan_ident,
+            self.position,
+        )
+
+
+@dataclass(frozen=True, kw_only=True)
+class AptMessage_MGMSG_MOT_REQ_POSCOUNTER(AptMessageHeaderOnlyNoParams):
+    message_id: ClassVar[AptMessageId] = AptMessageId.MGMSG_MOT_REQ_POSCOUNTER
```
auspicacious, line 799: This class actually does have a param -- there's a chan ident at position 2. You can inherit from `AptMessageHeaderOnlyChanIdent` to implement this.

src/pnpq/apt/protocol.py
```diff
@@ -688,6 +688,117 @@ class AptMessage_MGMSG_MOD_IDENTIFY(AptMessageHeaderOnlyChanIdent):
...
+            self.data_length,
+            self.destination_serialization,
+            self.source,
+            self.chan_ident,
+            self.position,
+        )
+
+
+@dataclass(frozen=True, kw_only=True)
+class AptMessage_MGMSG_MOT_SET_POSCOUNTER(AptMessageWithData):
```
auspicacious, line 745: Since the GET and SET command structures are essentially the same, we should avoid duplicating all this code and instead create a parent class that both can inherit from, containing all the serialization and deserialization logic. See AptMessageWithDataMotorStatus for an example of this.

tests/apt/test_protocol.py
```diff
@@ -215,6 +218,63 @@ def test_AptMessage_MGMSG_MOD_IDENTIFY_to_bytes() -> None:
...
+        position=200000,
+    )
+    assert msg.to_bytes() == bytes.fromhex("1004 0600 A2 01 0100400D0300")
+
+
+def test_AptMessage_MGMSG_MOT_REQ_POSCOUNTER_from_bytes() -> None:
+    msg = AptMessage_MGMSG_MOT_REQ_POSCOUNTER.from_bytes(b"\x11\x04\x00\x00\x50\x01")
+    assert msg.message_id == 0x0411
+    assert msg.destination == 0x50
+    assert msg.source == 0x01
```
auspicacious, line 268: You'll need to check for chan ident here.

tests/apt/test_protocol.py
```diff
@@ -215,6 +218,63 @@ def test_AptMessage_MGMSG_MOD_IDENTIFY_to_bytes() -> None:
     assert msg.to_bytes() == b"\x23\x02\x01\x00\x50\x01"
 
 
+def test_AptMessage_MGMSG_MOT_GET_POSCOUNTER_from_bytes() -> None:
+    msg = AptMessage_MGMSG_MOT_GET_POSCOUNTER.from_bytes(
+        bytes.fromhex("1204 0600 A2 01 0100400D0300")
+    )
+
+    assert msg.destination == 0x22
```
auspicacious, line 226: I see that there's an error in the official documentation where they say that `0xA2` should be channel 2 (`0x02`) after setting the highest bit to 0. Channel 2 would actually be transmitted as `0x82`.

It's fine to leave it as `0x22`, but it might be nice to leave a comment here about the error in the documentation.

src/pnpq/apt/protocol.py
auspicacious: I'm also not sure if it's necessary to explicitly type this as `ClassVar` here since it's already typed as such in the parent.

auspicacious (commented):
Thanks, please let me know if you have any questions.

tests/apt/test_protocol.py
master-of-ppap: To be honest, I am not too sure what you mean by this. To explain myself, the reason I assert `msg.destination == 0x22` is because I have `destination=Address(destination & 0x7F)` in the original function. And the reason I `& 0x'F` is because the documentation looked similar to `AptMessage_MGMSG_HW_GET_INFO` from which I based most of my code off of. 

I'll post your explanation in a comment for now I guess. I hope that's fine.

auspicacious: Your confusion is justified, as I was also confused. This was my mistake, although the documentation is also a problem.

On [page 63 of the protocol specification](https://www.thorlabs.com/Software/Motion%20Control/APT_Communications_Protocol.pdf) the destination in the example is called "Channel 2".

This is wrong, but not in the way that I thought it was wrong. I got confused.

It's wrong because, in the APT protocol, this is called an _address_, not a _channel_. An address describes what device is sending or receiving a message (source and destination). A channel describes which motor you are talking to on the device.

The documentation used the wrong term. It should say "Bay 1," instead of "Channel 2." [Bay 1's address is represented by `0x22`](https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/apt/protocol.py#L68). [Channel 2 is represented by `0x02`](https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/apt/protocol.py#L115).

I assume that the author of was counting from 1, not from 0, so what is called "Bay 1" in earlier parts of the specification became "Channel 2" here.

This is why it's important to make documentation precise and use vocabulary clearly and consistently.

As for why you are `& 0x7F`-ing the input value, read pages 34, 35, and 36 of the protocol specification.

tests/apt/test_protocol.py
```diff
@@ -215,6 +218,72 @@ def test_AptMessage_MGMSG_MOD_IDENTIFY_to_bytes() -> None:
...
+
+def test_AptMessage_MGMSG_MOT_SET_POSCOUNTER_from_bytes() -> None:
+    msg = AptMessage_MGMSG_MOT_SET_POSCOUNTER.from_bytes(
+        bytes.fromhex("1004 0600 A2 01 0100400D0300")
+    )
+
+    # There's an error in the official documentation
+    # where they say that 0xA2 should be channel 2 (0x02)
+    # after setting the highest bit to 0.
+    # Channel 2 would actually be transmitted as 0x82.
```
auspicacious, line 251:
```suggestion
# The official documentation's example refers to this as "Channel 2," but this is actually bay 1.
```

tests/apt/test_protocol.py
```diff
@@ -215,6 +218,72 @@ def test_AptMessage_MGMSG_MOD_IDENTIFY_to_bytes() -> None:
     assert msg.to_bytes() == b"\x23\x02\x01\x00\x50\x01"
 
 
+def test_AptMessage_MGMSG_MOT_GET_POSCOUNTER_from_bytes() -> None:
+    msg = AptMessage_MGMSG_MOT_GET_POSCOUNTER.from_bytes(
+        bytes.fromhex("1204 0600 A2 01 0100400D0300")
+    )
+
+    assert msg.destination == 0x22
```
auspicacious, line 226:
```suggestion
    # The official documentation's example refers to this as "Channel 2," but this is actually bay 1.
    assert msg.destination == 0x22
```

auspicacious (dismissed):
After you've understood what actually happened and updated the comments, you can merge this.

tests/apt/test_protocol.py
master-of-ppap: Yes. I understand now. But how come you are `& 0x7F`-ing instead of `| 0x80`-ing them? Wouldn't it have the same effect? Or, since the value is not strictly a 8bit integer, we are making sure that only the last seven least-significant bits are used. That makes sense.

Okay. I think I get it. I will merge now. Thank you.

tests/apt/test_protocol.py
master-of-ppap: Help. It seems like if I try to commit this, the tests don't pass anymore. Should I add this comment locally, run `black`, and re-commit it (from my own pc)?

auspicacious: In general, you should assume that I don't usually check any of my suggested code changes for correctness and they need to be tested yourself. In this case, I forgot the indentation.

`git pull` your branch locally to get the changes that were committed when you added this suggestion, reformat the code, commit and push it.

tests/apt/test_protocol.py
auspicacious: We do `| 0x80` the value before writing it out to bytes: https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/apt/protocol.py#L346

When reading in from bytes, we `& 0x7F` the value.

tests/apt/test_protocol.py
master-of-ppap: Oh yes. I see. That makes sense. 

Alright. I have done the deeds. Thank you as always for reviewing my code and providing very insightful comments. I definitely have learned a lot so far!

tests/apt/test_protocol.py
master-of-ppap: Ohhhhh. That makes a lot of sense. Since `bin(0x80) = '0b10000000'` and `bin(0x7F) '0b01111111'`, we set the most significant bit when we write it, and then we remove the most significant bit when we read it. That makes a lot of sense. Okay. Thank you.

auspicacious (approved):

PR #65, opened 2024-10-28
Refactor serial communication code into a new class

### LINKED ISSUE #29: Refactor MPC320 API and extract out common APT functionality
We have implemented the APT communication protocol for the Thorlabs MPC320.

The MPC320 is only one model of Thorlabs polarization controller, and there are several types of Thorlabs devices that use the APT communication protocol. We can generalize the control code to work with all kinds of devices.

## Acceptance criteria
* [Abstract base class](https://docs.python.org/3/library/abc.html) created for MPC type polarization controllers, in order to later support the MPC320 and MPC220 controllers
* `@abstractmethod` used to define the public interface of the device type
* APT communication code moved out of the MPC320 class, either to a common base class or to a utility class that can be called by implementations
* MPC320 class updated to use the moved APT code
* MPC320 implementation tested against the hardware

TomShen1234:
I'd say this closes #29 
A new `AptConnection` class was added, with the codes itself roughly unchanged from its initial implementation in the MPC320 class. 

To use the new class, initialize the connection object first, then pass it into the device class:

```
connection = AptConnection(serial_number="12345678")
return PolarizationControllerThorlabsMPC320(connection=connection)
```

### DIFF
--- a/src/pnpq/apt/connection.py
+++ b/src/pnpq/apt/connection.py
@@ -17,8 +17,6 @@ from .protocol import (
     Address,
     AptMessage,
     AptMessage_MGMSG_HW_REQ_INFO,
-    AptMessage_MGMSG_MOT_ACK_USTATUSUPDATE,
-    AptMessage_MGMSG_MOT_REQ_USTATUSUPDATE,
     AptMessageForStreamParsing,
     AptMessageId,
     ChanIdent,
@@ -80,15 +78,11 @@ class AptConnection:
         default_factory=threading.Lock
     )
 
-    tx_poller_thread: threading.Thread = field(init=False)
-    tx_poller_thread_lock: threading.Lock = field(default_factory=threading.Lock)
-
     log = structlog.get_logger()
 
     # Required inputs are defined below.
 
     serial_number: str
-
 
     def __post_init__(self) -> None:
         self.log.debug("Starting connection post-init...")
@@ -137,13 +131,6 @@ class AptConnection:
         #
         # TODO use some sort of thread manager to safely deal with
         # uncaught exceptions and other errors.
-        object.__setattr__(
-            self,
-            "tx_poller_thread",
-            threading.Thread(target=self.tx_poll, daemon=True),
-        )
-        self.tx_poller_thread.start()
-
         object.__setattr__(
             self,
             "rx_dispatcher_thread",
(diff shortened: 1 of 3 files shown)


src/pnpq/apt/connection.py
```diff
@@ -0,0 +1,358 @@
...
+                # TODO this is too general, do not catch Exception
+                except Exception as e:  # pylint: disable=W0718
+                    self.log.error(
+                        event=Event.UNCAUGHT_EXCEPTION,
+                        exc_info=e,
+                        partial_message=partial_message,
+                        full_message=full_message,
+                    )
+
+    def tx_poll(self) -> None:
```
auspicacious, line 238: So, by default, `tx_poll` probably shouldn't do anything. At least, it shouldn't send any messages.

So, we have a few options for how to solve this.

1. **Remove the `tx_poll` function, code, and thread completely from connection.py**: This is the easiest to do. We would move this part of the code into the device implementation classes.
2. **Provide a default no-op implementation in connection.py**: In this case, we would need to create a subclass of `AptConnection`, called, for instance, `StatusUpdatingAptConnection`, to implement polling as we think is appropriate.
3. **Pass in a `Callable` to the `AptConnection` constructor**: In other words, require a function to be passed in that `AptConnection` would run inside the `while True` loop, replacing lines 241 to 254.

Of the three options, I think that option 1 is probably the best. It gives us flexibility in what we can poll for and how we handle the results. One downside is that it means we have to do thread management in more places.

Option 2 gets tricky if we want to share information from the polling with the driver class.

Option 3 _could_ let us share information with the driver class, but in order to do so we would have to change the way that we initialize the connection and the driver object. And we would still have to deal with thread synchronization in the driver class.

auspicacious (commented):
@TomShen1234 This requires a little more work to make it generic enough for use across many different device models.

src/pnpq/apt/connection.py
TomShen1234: I see that different devices could implement polling slightly differently. But I thought the main point of this class is so that all the threading can happen separately. If we bring `tx_poll` back then we only abstracted 2/3 of the threading. It's definitely the quicker way to do this if we don't want to spend longer on this particular task tho.

TomShen1234:
Looks like black's complaining about an extra empty line after implementing the suggestions in Github...

src/pnpq/apt/connection.py
auspicacious: @TomShen1234 Another way of thinking about this is that the `AptConnection` class abstracts out the concept of an APT connection, e.g., the process of sending and receiving APT messages. It doesn't have anything to do with what messages should be sent using that connection.

For that matter, the fact that it uses threads is an implementation detail and not part of the public API.

Another consideration is that we might not always need polling. In theory, we can use `HW_START_UPDATEMSGS` to instead get the device to send messages to us automatically. The documentation on the `ACK` messages is also very confusing, and it's not clear if we really need them or not.

My inclination right now is still to make the polling part of the device driver class, not the connection class.

TomShen1234:
I moved it back. I'm wondering if the function should get a different name or become private with the `__` prefix. I presume it's not supposed to be called from the outside.

TomShen1234:
Just tested the code. It works fine on the real device.

auspicacious (approved):

PR #66, opened 2024-11-01
add 3 more messages: MOVE_JOG, MOVE_STOP, MOVE_STOPPED

master-of-ppap:
Partial implementation of #56 

Attempted to implement 3 more messages: MOVE_JOG, MOVE_STOP, MOVE_STOPPED
Created 2 new enums in the process: StopMode and JogDirection.

MOVE_JOG and MOVE_STOP use very similar code, so I am wondering if there's a better way to do this.

Also, while I have implemented the first 6 bytes of MSMSG_MOT_MOVE_STOPPED, the entire message is supposed to be 20 bytes. However, a similar message (MGMSG_MOT_MOVE_COMPLETED) is also in a similar format, but in reality the MPC320  does not follow up with the extra 14 bytes. So I have left it unfinished for now.

### DIFF
--- a/src/pnpq/apt/protocol.py
+++ b/src/pnpq/apt/protocol.py
@@ -146,8 +146,8 @@ class StopMode(int, Enum):
 class JogDirection(int, Enum):
     """Used in MSMSG_MOT_MOVE_JOG."""
 
-    JOG_FORWARD = 0x01
-    JOG_BACKWARD = 0x02
+   FORWARD = 0x01
+   REVERSE = 0x02
 
 
 @enum.unique
--- a/src/pnpq/apt/protocol.py
+++ b/src/pnpq/apt/protocol.py
@@ -146,8 +146,8 @@ class StopMode(int, Enum):
 class JogDirection(int, Enum):
     """Used in MSMSG_MOT_MOVE_JOG."""
 
-   FORWARD = 0x01
-   REVERSE = 0x02
+    FORWARD = 0x01
+    REVERSE = 0x02
 
 
 @enum.unique
--- a/src/pnpq/apt/protocol.py
+++ b/src/pnpq/apt/protocol.py
@@ -45,6 +45,10 @@ class AptMessageId(int, Enum):
     MGMSG_POL_REQ_PARAMS = 0x0531
     MGMSG_POL_SET_PARAMS = 0x0530
 
+    MGMSG_MOT_MOVE_JOG = 0x046A
+    MGMSG_MOT_MOVE_STOP = 0x0465
+    MGMSG_MOT_MOVE_STOPPED = 0x0466
+
     MGMSG_RESTOREFACTORYSETTINGS = 0x0686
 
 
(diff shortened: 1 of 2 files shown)


src/pnpq/apt/protocol.py
```diff
@@ -864,6 +880,86 @@ class AptMessage_MGMSG_POL_SET_PARAMS(AptMessageWithDataPolParams):
...
+@dataclass(frozen=True, kw_only=True)
+class AptMessage_MGMSG_MOT_MOVE_STOPPED(AptMessageHeaderOnlyChanIdent):
+    """Note that the APT documentation indicates that this should be
+    followed by a full USTATUS data packet.
+
+    MGMSG_MOT_MOVE_COMPLETED is a similar message that is also expected to be
+    followed by a USTATUS packet. For MOVE_COMPLETED, in reality no data packet
+    follows for the MPC320, so further testing is required if this is also true
+    for MOVE_STOPPED.
+    """
```
auspicacious, line 958:
```suggestion
    """Note that the APT documentation indicates that this should be
    followed by a full USTATUS data packet. In reality, for the
    MPC320, no data packet follows."""
```

src/pnpq/apt/protocol.py
```diff
@@ -134,6 +134,22 @@ def from_bool(cls, toggle: bool) -> "EnableState":
...
+    IMMEDIATE_STOP = 0x01
+    CONTROLLED_STOP = 0x02
+
+
+@enum.unique
+class JogDirection(int, Enum):
+    """Used in MSMSG_MOT_MOVE_JOG."""
+
+    JOG_FORWARD = 0x01
+    JOG_BACKWARD = 0x02
```
auspicacious, line 150: Removing the redundant "JOG" and using the same words as the specification ("reverse" instead of "backward," not a big deal, but "reverse" is used on cars, too, so it's probably just as good). It's probably best not to merge this suggestion but update it on your machine so you can update any references to it.

```suggestion
   FORWARD = 0x01
   REVERSE = 0x02
```

src/pnpq/apt/protocol.py
```diff
@@ -134,6 +134,22 @@ def from_bool(cls, toggle: bool) -> "EnableState":
         return cls.CHANNEL_DISABLED
 
 
+@enum.unique
+class StopMode(int, Enum):
+    """Used in MSMSG_MOT_MOVE_STOP."""
+
+    IMMEDIATE_STOP = 0x01
+    CONTROLLED_STOP = 0x02
```
auspicacious, line 142: Removing the redundant "STOP." It's probably best not to merge this suggestion but update it on your machine so that you can update references to it at the same time.

```suggestion
    IMMEDIATE = 0x01
    CONTROLLED = 0x02
```

auspicacious (commented):
Thanks to Friday's testing we know that `MOVE_JOG` and `MOVE_STOPPED` work, and I'm not too concerned about `MOVE_STOP` -- that's a low priority for us to implement. Once these small changes are made, it should be good to merge.

auspicacious (approved):
This looks complete, so I'm approving it.

FYI after you push PR changes, it's a good idea to leave a comment asking me to re-review. Otherwise I just see a bunch of notifications about new changes being pushed and I don't know if you're finished yet.

master-of-ppap:
Oh yes. That's a good idea. I should probably do that. Thank you for the review.

PR #67, opened 2024-11-04
Add sleep after sending no-reply commands.

auspicacious:
One-line code change with a lot of explanatory comments.

@TomShen1234 @master-of-ppap I think that we should merge this PR, then, after @master-of-ppap finishes writing the driver implementation for `jog` in a separate PR, we should re-test on the hardware.

I am a little concerned that this will add 0.4 seconds to every move command. We might be able to tune this downward a little. But we have to disable the channel after each move if we want to use the device accurately, since otherwise it will just sit there vibrating, and that means having some kind of short delay.

### DIFF
--- a/src/pnpq/apt/connection.py
+++ b/src/pnpq/apt/connection.py
@@ -236,6 +236,27 @@ class AptConnection:
                 if match_reply is None:
                     with self.tx_connection_lock:
                         self.connection.write(message.to_bytes())
+                        # Some no-reply commands take time to
+                        # complete. Sending other messages while this
+                        # is happening could cause the device's
+                        # internal software to fail until a hard reset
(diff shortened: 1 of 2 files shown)


master-of-ppap:
Sounds good. Should I open a new PR with my jogging code from last Friday? And clean it up a bit ofc.

auspicacious:
@master-of-ppap Yes, clean up the jog code and make a PR. And I would appreciate it if you could approve this pull request.

TomShen1234 (approved):
Looks fine to me. I'll approve it. 
I guess in the near future we could do a test to see whether the delay could be shortened (or check with the experimenters to see whether it needs to be shortened or not).

PR #72, opened 2024-11-05
Changed MPC320 parameters to TypedDict

### LINKED ISSUE #68: Refactor MPC320 device parameters functionality
The implementation of [the MPC320's polarization controller params](https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/devices/polarization_controller_thorlabs_mpc320.py#L30) commands probably causes more problems than it solves.

For one, it maintains a local cache of the state of the params as last seen, which is initialized at startup. It is likely easier and better for us to simply require users to query the device directly each time they wish to know the state of the params.

For another, the method for setting the params, while convenient, causes problems with type checking.

Ideally, we should be able to continue using this convenient method for setting params, while having strong type checking. This may be possible if we turn the PolarizationControllerParams class into a [TypedDict](https://typing.readthedocs.io/en/latest/spec/typeddict.html#typeddict) -- which might be possible if we simply remove the default values.

If not, we should at least make sure that there is a path for users to be able to set the value of only one parameter at a time, without changing the other parameters' values.

TomShen1234:
Closes #68 

I have yet to test this code on the real device. It would be nice if I can have some support on that. 

I removed the `dataclass` decorator on `PolarizationControllerParams`, and changed it into a `TypedDict`. The class property is also removed in favor of `refresh_params` returning the dictionary. 

The only change is that the parameter now have to be used with the dictionary syntax rather than a class (which requires some string accessed variables). Let's discuss whether this is ideal. 

The interface for `set_params` remains unchanged, with the underlying logic changed to fetch the current params, update the dictionary, and sending the updated dictionary back to the device.

### DIFF
--- a/src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
+++ b/src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
@@ -226,7 +226,7 @@ class PolarizationControllerThorlabsMPC320:
                 destination=Address.GENERIC_USB,
                 source=Address.HOST_CONTROLLER,
                 velocity=params["velocity"],
-                home_position=params["home_position"].magnitude,
+                home_position=round(params["home_position"].magnitude),
                 jog_step_1=params["jog_step_1"],
                 jog_step_2=params["jog_step_2"],
                 jog_step_3=params["jog_step_3"],
--- a/src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
+++ b/src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
@@ -1,7 +1,7 @@
-import dataclasses
 import threading
 import time
 from dataclasses import dataclass, field
+from typing import TypedDict
 
 import structlog
 from pint import Quantity
@@ -26,13 +26,12 @@ from ..apt.protocol import (
 from ..units import ureg
 
 
-@dataclass(kw_only=True)
-class PolarizationControllerParams:
-    velocity: int = 0
-    home_position: Quantity = 0 * ureg.degree
-    jog_step_1: int = 0
-    jog_step_2: int = 0
-    jog_step_3: int = 0
+class PolarizationControllerParams(TypedDict):
+    velocity: int
+    home_position: Quantity
+    jog_step_1: int
+    jog_step_2: int
+    jog_step_3: int
 
 
 @dataclass(frozen=True, kw_only=True)
@@ -52,12 +51,6 @@ class PolarizationControllerThorlabsMPC320:
             ChanIdent.CHANNEL_2,
             ChanIdent.CHANNEL_3,
         ]
-    )
-
-    # Stored in a non-frozen dataclass so that we can refresh them as
-    # the configuration changes
-    params: PolarizationControllerParams = field(
-        default_factory=PolarizationControllerParams
     )
 
     def __post_init__(self) -> None:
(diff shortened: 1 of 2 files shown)


src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
```diff
@@ -169,7 +162,7 @@ def move_absolute(self, chan_ident: ChanIdent, position: Quantity) -> None:
         self.log.debug("move_absolute command finished", elapsed_time=elapsed_time)
         self.set_channel_enabled(chan_ident, False)
 
-    def refresh_params(self) -> None:
+    def refresh_params(self) -> PolarizationControllerParams:
```
auspicacious, line 165: Refactor this to be called `get_params`

src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
```diff
@@ -213,28 +209,29 @@ def set_params(
...
-                jog_step_2=new_params.jog_step_2,
-                jog_step_3=new_params.jog_step_3,
+                velocity=params["velocity"],
+                home_position=params["home_position"].magnitude,
+                jog_step_1=params["jog_step_1"],
+                jog_step_2=params["jog_step_2"],
+                jog_step_3=params["jog_step_3"],
             )
         )
         time.sleep(1)
```
auspicacious, line 237: We can remove this `sleep` now that we're sleeping on every no-reply command in `connection.py`.

auspicacious (commented):
I think this approach is good enough, passing around dictionaries is a very Pythonic way of doing things, we get type safety this way, and we are unlikely to need to add methods, so a dataclass isn't necessary.

Just a couple of small changes and I think it should be good to go. Could you maybe also include a test to exercise this functionality, maybe by adjusting the home position to some arbitrary value, confirming that homing goes to the new position, and then resetting home back to zero?

src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
```diff
@@ -213,28 +207,28 @@ def set_params(
...
             AptMessage_MGMSG_POL_SET_PARAMS(
                 destination=Address.GENERIC_USB,
                 source=Address.HOST_CONTROLLER,
-                velocity=new_params.velocity,
-                home_position=new_params.home_position.magnitude,
-                jog_step_1=new_params.jog_step_1,
-                jog_step_2=new_params.jog_step_2,
-                jog_step_3=new_params.jog_step_3,
+                velocity=params["velocity"],
+                home_position=params["home_position"].magnitude,
```
auspicacious, line 229: I think this might be a bug.

I don't see where we are doing the conversion of the Quantity into encoder counts (steps) before sending the message.

src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
```diff
@@ -213,28 +207,28 @@ def set_params(
...
```
auspicacious, line 229: I've realized there's a serious mismatch between step units, which are discrete, and degrees/radians, which are continuous. It almost makes sense to try and solve this at the time of unit conversion, but at a glance I don't see anything in the Pint library that allows rounding.
```suggestion
                home_position=round(params["home_position"].magnitude),
```

auspicacious (approved):

PR #73, opened 2024-11-07
implement class and tests for AptMessage_MGMSG_MOT_SET_EEPROMPARAMS

master-of-ppap:
Implementation for #56 

* Implement class and tests for last message for MPC320 `AptMessage_MGMSG_MOT_SET_EEPROMPARAMS`

### DIFF
--- a/src/pnpq/apt/protocol.py
+++ b/src/pnpq/apt/protocol.py
@@ -961,7 +961,7 @@ class AptMessage_MGMSG_MOT_SET_EEPROMPARAMS(AptMessageWithData):
     )
 
     chan_ident: ChanIdent
-    specified_msg_id: int
+    message_id_to_save: AptMessageId
 
     @classmethod
     def from_bytes(cls, raw: bytes) -> "AptMessage_MGMSG_MOT_SET_EEPROMPARAMS":
@@ -971,7 +971,7 @@ class AptMessage_MGMSG_MOT_SET_EEPROMPARAMS(AptMessageWithData):
             destination,
             source,
             chan_ident,
-            specified_msg_id,
+            message_id_to_save,
         ) = cls.message_struct.unpack(raw)
 
         if message_id != cls.message_id:
@@ -991,7 +991,7 @@ class AptMessage_MGMSG_MOT_SET_EEPROMPARAMS(AptMessageWithData):
             destination=Address(destination & 0x7F),
             source=Address(source),
             chan_ident=ChanIdent(chan_ident),
-            specified_msg_id=specified_msg_id,
+            message_id_to_save=message_id_to_save,
         )
 
     def to_bytes(self) -> bytes:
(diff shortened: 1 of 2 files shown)


src/pnpq/apt/protocol.py
```diff
@@ -958,3 +950,56 @@ class AptMessage_MGMSG_MOT_MOVE_STOPPED(AptMessageHeaderOnlyChanIdent):
...
+@dataclass(frozen=True, kw_only=True)
+class AptMessage_MGMSG_MOT_SET_EEPROMPARAMS(AptMessageWithData):
+    data_length: ClassVar[int] = 4
+    message_id = AptMessageId.MGMSG_MOT_SET_EEPROMPARAMS
+    message_struct: ClassVar[Struct] = Struct(
+        f"{AptMessageWithData.header_struct_str}{ATS.WORD}{ATS.WORD}"
+    )
+
+    chan_ident: ChanIdent
+    specified_msg_id: int
```
auspicacious, line 964: This should be type `AptMessageId`

auspicacious, line 964: `message_id_to_save`

master-of-ppap:
Thank you. I have implemented those changes.

However, changing the type to `AptMessageId` creates some typing issues with the test code:
```py
msg = AptMessage_MGMSG_MOT_SET_EEPROMPARAMS(
        destination=Address.GENERIC_USB,
        source=Address.HOST_CONTROLLER,
        chan_ident=ChanIdent.CHANNEL_1,
        message_id_to_save=0x04B6, # message_id_to_save is type AptMessageId, but 0x04B6 is type int
    )
```
As we are trying to save type `int` to type `AptMessageId`. I am not sure what the best workaround for this problem is, since we haven't implemented any of the other messages that `SET_EEPROMPARAMS` can save. So, for now, I have just used a random message that we have implemented and inserted a little comment in the code. Would this work?

src/pnpq/apt/protocol.py
```diff
@@ -958,3 +950,56 @@ class AptMessage_MGMSG_MOT_MOVE_STOPPED(AptMessageHeaderOnlyChanIdent):
...
+        if destination & 0x80 != 0x80:
+            raise ValueError(
+                f"Expected the destination's highest bit to be 1, indicating that a data packet follows, but it was 0. Full raw data was {raw!r}"
+            )
+
+        return AptMessage_MGMSG_MOT_SET_EEPROMPARAMS(
+            destination=Address(destination & 0x7F),
+            source=Address(source),
+            chan_ident=ChanIdent(chan_ident),
+            message_id_to_save=message_id_to_save,
```
auspicacious, line 994: There's an ugly little bug here that should have been caught by type checking but wasn't. I think this is because I made the enum inherit from `int`. I'm rethinking that decision, but I can't change that as part of this PR.

```suggestion
            message_id_to_save=AptMessageId(message_id_to_save),
```

After making this change, one of your tests will start to fail.

master-of-ppap:
Yes, I see what you mean about that ugly little bug. It wouldn't be possible for all `int`s to be mapped to `AptMessageId`s, so that is a potential cause for concern.

Anyway, to fix the check I made it so both methods now use `MGMSG_HW_GET_INFO`. It should be good for now, as these are just unit tests for the classes.

auspicacious (approved):
OK, this looks good enough.

PR #75, opened 2024-11-08
Jog driver mpc320

### LINKED ISSUE #70: Implement MPC320 jog functionality
Implement a jog function in the MPC320 driver, so that it is possible to repeatedly and easily move the device's paddles forward or backward in small steps.

The jog distance is configured separately, in the device parameters, and should not be settable through this function.

## Acceptance criteria
* Docstring for the function indicating where to set the jog distance
* Motor channels are enabled before the jog and disabled after the jog, to prevent motor vibration
* Tested against actual device

master-of-ppap:
Closes #70 

Added support for jog functionality for MPC320, allowing the user to jog real devices.
* Added docstring
* Motor channels are enabled and disabled when performing jog 
* Tested against actual device (one week ago)

### DIFF
--- a/hardware_tests/test_polarization_controller_thorlabs_mpc320.py
+++ b/hardware_tests/test_polarization_controller_thorlabs_mpc320.py
@@ -30,13 +30,29 @@ def test_move_absolute(device: PolarizationControllerThorlabsMPC320) -> None:
     device.move_absolute(ChanIdent.CHANNEL_2, 160 * ureg.degree)
     device.move_absolute(ChanIdent.CHANNEL_3, 160 * ureg.degree)
 
+        # device.move_absolute(ChanIdent.CHANNEL_2, 30 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_1, 90 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_3, 90 * ureg.degree)
+
+    # device.move_absolute(ChanIdent.CHANNEL_3, 165 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_2, 90 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_1, 0 * ureg.degree)
+
+    # device.move_absolute(ChanIdent.CHANNEL_1, 10 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_1, 100 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_1, 50 * ureg.degree)
+
+    device.home(ChanIdent.CHANNEL_1)
+    device.home(ChanIdent.CHANNEL_2)
+    device.home(ChanIdent.CHANNEL_3)
+
     # One of the channels on our test device appears to forget to turn
     # off its motor when it's homed or set to 0 degrees. It just sits
     # there vibrating and whining. It's not really safe to leave the
     # device at degree 0 for this reason. 170 also seems too far (160 seems about the safest)
-    device.move_absolute(ChanIdent.CHANNEL_1, 10 * ureg.degree)
-    device.move_absolute(ChanIdent.CHANNEL_2, 10 * ureg.degree)
-    device.move_absolute(ChanIdent.CHANNEL_3, 10 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_1, 10 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_2, 10 * ureg.degree)
+    # device.move_absolute(ChanIdent.CHANNEL_3, 10 * ureg.degree)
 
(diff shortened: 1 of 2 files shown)


hardware_tests/test_polarization_controller_thorlabs_mpc320.py
```diff
@@ -75,6 +75,25 @@ def test_move_absolute(device: PolarizationControllerThorlabsMPC320) -> None:
     # device.move_absolute(ChanIdent.CHANNEL_3, 0 * ureg.degree)
 
 
+def test_jog(device: PolarizationControllerThorlabsMPC320) -> None:
+
+    jog_step_1 = 50
+    jog_count_1 = 5
+
+    device.set_params(jog_step_1=jog_step_1, jog_step_2=25, jog_step_3=25)
```
auspicacious, line 83: You can just set `jog_step_1`, you don't need to touch the other two.
```suggestion
    device.set_params(jog_step_1=jog_step_1)
```

hardware_tests/test_polarization_controller_thorlabs_mpc320.py
```diff
@@ -75,6 +75,25 @@ def test_move_absolute(device: PolarizationControllerThorlabsMPC320) -> None:
...
+
+    device.home(ChanIdent.CHANNEL_1)
+    device.home(ChanIdent.CHANNEL_2)
+    device.home(ChanIdent.CHANNEL_3)
+
+    # Compare jog distance versus absolute distance
```
auspicacious, line 89: It's best to spell out exactly what is going on here and what we expect. Something like "Validate that we jogged to the expected position. If we are at the correct position, the move_absolute command should not cause the device to move at all." just above line 92.

hardware_tests/test_polarization_controller_thorlabs_mpc320.py
```diff
@@ -75,6 +75,25 @@ def test_move_absolute(device: PolarizationControllerThorlabsMPC320) -> None:
...
+    for _ in range(jog_count_1):
+        device.jog(ChanIdent.CHANNEL_1, JogDirection.FORWARD)
+    device.move_absolute(
+        ChanIdent.CHANNEL_1, jog_count_1 * jog_step_1 * ureg.mpc320_step
+    )
```
auspicacious, line 94: It would be good if we reset the `jog_step_1` param after running this.

After tomshen's PR to update how getting and setting params works is merged, we can do a `get_params` at the beginning of the test, store the value of `jog_count_1`, and then `set_params` at the end to reset it.

Even better, we can do this using `try` and `finally`, so that we can (almost) guarantee that the params value is reset even if something fails in the `jog` command.

src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
```diff
@@ -139,6 +142,30 @@ def identify(self, chan_ident: ChanIdent) -> None:
             )
         )
 
+    def jog(self, chan_ident: ChanIdent, jog_direction: JogDirection) -> None:
+        """
+        Jogs the device forward or backwards in small steps.
+
+        The specific amount of steps per jog can be set via the PolarizationcontrollerThorlabsMPC320.set_params() function.
```
auspicacious, line 149: I'm going to add Sphinx documentation generation in a PR soon. This would be a good place to try out the ability to actually make a hyperlink to the `set_params` function. Stay tuned.

hardware_tests/test_polarization_controller_thorlabs_mpc320.py
master-of-ppap: Yes, that sounds like a good idea. However, if I recall correctly, I remember you said something about not going too overboard with resetting the params of the machine, because there's a limited amount of writes or something. Or maybe I am misremembering.

src/pnpq/devices/polarization_controller_thorlabs_mpc320.py
master-of-ppap: Ooh. Can't wait for the Sphinx!

master-of-ppap:
Okay, for now I'll wait for Tom's PR to be merged I think.

hardware_tests/test_polarization_controller_thorlabs_mpc320.py
auspicacious: Since we run these tests manually, they will only be run a small number of times.

We don't know how the storage is implemented internally -- it could be an EEPROM or Flash storage -- and we don't know the lifetime of the storage. I've heard numbers ranging from 100,000 to 1,000,000 write cycles for EEPROM to much larger for Flash. We probably don't need to worry about it for now.

It might be that we should be implementing jog in software, in the end.

master-of-ppap: Oh yes I see. Yeah, implementing it in software would probably be a good idea either way.

auspicacious (approved):

Issue #71, opened 2024-11-05
Implement basic Sphinx documentation


auspicacious:
As our open projects mature, we need to begin documenting them properly. We will use [Sphinx](https://www.sphinx-doc.org/en/master/) and its [Autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#module-sphinx.ext.autodoc) extension to do this. Using the default reStructuredText approach instead of simpler approaches such as the Google Python documentation format or Markdown-based formats will ensure that we have the flexibility and experience to make documentation as complex and powerful as we would like.

## Acceptance criteria
* Sphinx HTML documentation generation works
* API documentation is generated
* Generation is performed in the `check.bash` script with `-W, --fail-on-warning` and `-E, --fresh-env` enabled to detect problems
* At least one function is fully documented using appropriate RST tags so that we have an example to use elsewhere

PR #77, opened 2024-11-11
Add Sphinx documentation, automatically generate API docs

auspicacious:
Closes https://github.com/moonshot-nagayama-pj/PnPQ/issues/71

I'm still not 100% happy with the output formatting, but this is a start.

If you would like to see the output, check out this branch, run `bin/check.bash`, and then open the file `sphinx/build/html/index.html`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,10 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+sphinx/build
+
+# API documents are auto-generated
+sphinx/source/api
 
 # PyBuilder
 .pybuilder/
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ PnPQ is a python library package for controlling optical devices in quantum netw
 
 ## Prerequisites
 
-- (rye)[https://rye.astral.sh/].
+- [rye](https://rye.astral.sh/).
 - shellcheck
 - shfmt
 
--- a/bin/check.bash
+++ b/bin/check.bash
@@ -68,6 +68,9 @@ shfmt --diff --simplify .
 stdmsg "Running rye lint..."
 rye lint
 
+stdmsg "Building documentation..."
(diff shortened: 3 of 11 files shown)


auspicacious:
That's odd, the check script worked locally. Looking into it now.

auspicacious:
That's fixed. Sphinx was complaining about a missing directory.

TomShen1234 (commented):
It works, but how should we host this site? 
Maybe host it in Github Page via a worker?
Apart from that (and the doc being slightly messy and hard to read), I think it's good.

auspicacious:
@TomShen1234 See https://github.com/moonshot-nagayama-pj/PnPQ/issues/78 for publishing. I would like to keep this PR scoped to the initial generation process, especially since we don't have a lot of documentation to publish yet.

Could you approve this PR if there's no other comments?

TomShen1234 (approved):
Alright

master-of-ppap:
This looks very nice. This will be very helpful for sure. Should we start updating the docstrings of some functions to hyperlinks like you said previously?

auspicacious:
@master-of-ppap I did update the one, I'm not sure how many others exist: https://github.com/moonshot-nagayama-pj/PnPQ/pull/77/files#diff-b957e7273e2cb620ca5b5e58848356de109c0d8d8854553cf6caac5a28c951fbR141

In general, let's start thinking more about documentation, especially as we start to implement more driver classes.

master-of-ppap:
@auspicacious Oh yes nice, but I think there are still some more around the code.
For example, on line 137 in protocol.py
```py
@enum.unique
class JogDirection(int, Enum):
    """Used in MSMSG_MOT_MOVE_JOG."""

    FORWARD = 0x01
    REVERSE = 0x02
```

and also the stopmode enum above also could be converted into links as well.
This is gonna be very useful I can tell.

PR #80, opened 2024-11-15
Initial log rotation support

### LINKED ISSUE #69: Implement log rotation for test logs
We are learning how to use [structlog](https://www.structlog.org/en/stable/). In particular, we will need to know how to configure it, so that when it is integrated into our services and applications, we have a good example of a configuration ready to use.

Moreover, our debug-level logging can be quite verbose, and log files can increase in size quickly.

Improve our test log configuration, in `hardware_tests/conftest.py` and `tests/conftest.py`, to use log rotation.

## Acceptance criteria
* Log files are rotated after some criterion is met, either maximum size, time/date, or both.
* Old log files are deleted after some threshold, for instance, after there are 5 old log files.
* Ideally, the two separate `conftest.py` files are updated to pull configuration from the same place, so that we no longer need to maintain two separate configurations.

TomShen1234:
Closes #69 

The structlog code was updated to wrap the system logger, which is configured to use the `RotatingFileHandler`.

Currently, it's configured to rotate every 100MB, and keeps the last 3 file (in addition to the current file). I have a feeling that this has to change, let me know the desired value. 

Apart from this, the existing stub devices are also updated to support structlog. I have a feeling that the waveplate stub would have to be refactored in the near future, so I didn't create a dedicated `Event` for every action, only the most basic ones (the other ones are still string logs).


src/pnpq/events.py
```diff
@@ -9,3 +9,16 @@ class Event(StrEnum):
     TX_MESSAGE_ORDERED = auto()
     TX_MESSAGE_UNORDERED = auto()
     UNCAUGHT_EXCEPTION = auto()
+
+    # General hardware errors
+    DEVICE_CONNECTED = auto()
+    DEVICE_NOT_CONNECTED_ERROR = auto()
```
auspicacious, line 15: For errors, I'm of two minds. One is to remove `_ERROR` completely and rely on the log level to identify errors. The other is to move `ERROR` to the front, so that this is `ERROR_DEVICE_NOT_CONNECTED`. Why? Because it's faster to match with a regex, for one thing, or if you're reading code and looking for errors.

Thoughts?

tests/logs.py
```diff
@@ -0,0 +1,41 @@
...
+
+def find_project_dir(path: Path) -> Path:
+    if path.is_dir():
+        for contained_path in path.iterdir():
+            if contained_path.name == "pyproject.toml":
+                return path
+    return find_project_dir(path.parent)
+
+
+def setup_log(log_file_name: str) -> None:
```
auspicacious, line 16: This mostly follows [the advice in structlog's docs](https://www.structlog.org/en/stable/standard-library.html) and it looks pretty good, but it would be really nice if all logs, including logs written to the standard logger, were formatted as JSON consistently.

Can you go farther, [look at this section of the docs](https://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging), and make sure that everything is logged in JSON format? This could be useful if we need to debug behavior in third-party libraries we don't control.

auspicacious (commented):
I think this configuration is good enough and your current log rotation parameters are probably OK.

When we're running production services, we will have to choose between writing everything to stdout and having systemd handle writing those logs to the journal and shipping them to our log aggregator (so we wouldn't need to do log rotation at all), or writing them into a ramdisk and having the aggregator pick them up from there (where 1GB of logs might be OK). And we can always tweak the parameters. This looks good for testing.

In the future, we will also want to set certain information in the logging context alongside the events so that we know which device is doing what action, for example. This is a good step in that direction.

That being said, it would be nice to get everything JSON-formatted now, as I mentioned in one of the comments.

src/pnpq/events.py
TomShen1234: Good point. Given that the `UNCAUGHT_EXCEPTION` event didn't have the `ERROR`, maybe I'll remove it to keep consistency.

TomShen1234:
The code has been updated. There's 2 more things that we should decide whether we need or not:
- `cache_logger_on_first_use=True` flag - I guess this could make the code slightly more performant...
- `structlog.stdlib.ProcessorFormatter.remove_processors_meta` processor - Currently every log record contains a `"_record"` key containing a file path, line number, and event name of the log. Although the information could be useful, but it makes each logs almost twice as long. There's also a `"_from_structlog"` boolean that's true if the `structlog` logger is used to log it. This processors removes both of them.

auspicacious:
@TomShen1234

* Yes, cache loggers
* Remove both of the `_` properties. It's good to know that this metadata is available if we want to add it into the log object selectively in the future, but we don't need this information in every log line.

Thanks, it otherwise looks complete.

TomShen1234:
I updated the configurations. 
Also, the PR was edited to close the original issue.

auspicacious (approved):
Thank you, approved.

PR #82, opened 2024-11-22
Adopt uv

### LINKED ISSUE #81: Refactor package manager to uv
Since rye is becoming deprecated, we should replace it with the recommended replacement: [uv](https://docs.astral.sh/uv/).

## Acceptance Criteria

- `pyproject.toml` updated to support `uv`
- Check scripts are updated accordingly

TomShen1234:
Closes #81 

The project's updated to use the `uv` package manager instead of `rye`. The check scripts are also updated accordingly. 

Several things needs to be adjusted:

- [Install `uv`](https://docs.astral.sh/uv/getting-started/installation/)
- Instead of `rye sync`, now `uv sync` should be run
- It would be ideal to recreate the virtual environment in all local copies


TomShen1234:
I'm back. 
I just checked the compatibility using the testbed-node repository. It does sync fine.

PR #84, opened 2024-12-02
Mock controller to test mpc320's move_absolute()

### LINKED ISSUE #74: Basic implementation of mock connection object
In order to add unit tests our device drivers, we would need a mock for the connection class that can read the messages sent by the drivers, and send the corresponding response. It should behave the same way and have the same interface as the current `AptConnection` class. 

To start this task, we should keep it simple, and support just the `move_absolute` function of the MPC320. 

This task should also let us decide what library should be used for the mock. 

## Acceptance criteria
- A mock for `AptConnection` created.
- It should accept the `AptMessage_MGMSG_MOT_MOVE_ABSOLUTE` and return `AptMessage_MGMSG_MOT_GET_USTATUSUPDATE`, like the real device. 
- Unit test should be created for the `move_absolute` function of the driver.

master-of-ppap:
Closes #74. In theory.

Additions: 
* Created a new test file called `test_mock.py` (is this a good file name?). 
* Uses the `Mock` object from the `unittest` library to mock the behavior of the `AptConnection` class.
* Supports only the `move_absolute` function of the `PolarizationControllerThorlabsMPC320` at the moment.
* Uses the sideeffect functionality to simulate the driver accepting a message, and returning a hypothetical reply. The unit test then checks if whether or not the driver correctly recognizes this hypothetical reply. 
* Asserts that the `AptConnection.send_message_expect_reply()` was correctly called three times. (Two for enabling and disabling the channel, one for the movement itself)

Considerations:
* I would like to assert if the `AptMessage_MGMSG_MOT_MOVE_ABSOLUTE` message was correctly sent while the channel was enabled, but I'm not sure how to do that.
* Defining tests for every single driver function would resulting in having to define a new hypothetical reply message each time. So, this may not be the most efficient way of doing these tests.

### DIFF
--- a/tests/test_mock.py
+++ b/tests/test_mock.py
@@ -1,0 +1,58 @@
+from typing import Callable
+from unittest.mock import Mock, create_autospec
+
+from pnpq.apt.connection import AptConnection
+from pnpq.apt.protocol import (
+    Address,
+    AptMessage,
+    AptMessage_MGMSG_MOT_GET_USTATUSUPDATE,
+    AptMessage_MGMSG_MOT_MOVE_ABSOLUTE,
+    ChanIdent,
+    Status,
+    StatusBits,
+)
+from pnpq.devices.polarization_controller_thorlabs_mpc320 import (
+    PolarizationControllerThorlabsMPC320,
+)
+from pnpq.units import pnpq_ureg
+
+
+def test_polarization_controller_move_absolute() -> None:
+
+    connection = create_autospec(AptConnection)
+
+    def mock_send_message_expect_reply(
+        sent_message: AptMessage,
+        match_reply_callback: Callable[
+            [
+                AptMessage,
+            ],
+            bool,
+        ],
+    ) -> None:
+        # Only send a reply if the message is a MOVE_ABSOLUTE message
+        if isinstance(sent_message, AptMessage_MGMSG_MOT_MOVE_ABSOLUTE):
+
+            # A hypothetical reply message from the device
+            reply_message = AptMessage_MGMSG_MOT_GET_USTATUSUPDATE(
+                chan_ident=sent_message.chan_ident,
+                position=sent_message.absolute_distance,
+                velocity=50,
+                motor_current=3 * pnpq_ureg.milliamp,
+                status=Status.from_bits(StatusBits.ACTIVE),
+                destination=Address.HOST_CONTROLLER,
+                source=Address.GENERIC_USB,
+            )
+
+            assert match_reply_callback(reply_message)
+
+    connection.send_message_expect_reply.side_effect = mock_send_message_expect_reply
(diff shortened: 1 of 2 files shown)


tests/test_mock.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def mock_send_message_expect_reply(
+        sent_message: AptMessage,
+        match_reply_callback: Callable[
+            [
+                AptMessage,
+            ],
+            bool,
+        ],
+    ) -> None:
```
auspicacious, line 32: I'm somewhat surprised that mypy is OK with this having a different return type than the real function, but this is good enough for now.

tests/test_mock.py
```diff
@@ -0,0 +1,58 @@
...
+        # Only send a reply if the message is a MOVE_ABSOLUTE message
+        if isinstance(sent_message, AptMessage_MGMSG_MOT_MOVE_ABSOLUTE):
+
+            # A hypothetical reply message from the device
+            reply_message = AptMessage_MGMSG_MOT_GET_USTATUSUPDATE(
+                chan_ident=sent_message.chan_ident,
+                position=sent_message.absolute_distance,
```
auspicacious, line 39: We need to ask what we're really testing... with this design, we don't have any evidence that the correct values were sent. It would be better if we actually asserted that the `chan_ident` is 1 and the position is as expected on the `sent_message` before using those values to construct the `reply_message`.

auspicacious (commented):
We're pretty close!

> Created a new test file called test_mock.py (is this a good file name?).

This file is specifically for testing the MPC320 driver, so it should be named after the driver file, e.g. `polarization_controller_thorlabs_mpc320.py`.

> Defining tests for every single driver function would resulting in having to define a new hypothetical reply message each time. So, this may not be the most efficient way of doing these tests.

We probably need a different setup for each test anyway, so it's not that big of a deal. As we move forward, we can try to see if there's any improvements we can make that will make things less verbose.

tests/test_mock.py
master-of-ppap: We don't have any return statement for this function though. Wouldn't that mean it returns `None`?

master-of-ppap:
Thanks for the feedback. I have made the changes.

tests/test_mock.py
auspicacious: Yeah, but I'm thinking about this from the point of view of a strongly typed language. If we're implementing the interface defined by the `AptConnection` class, then this function has to conform to the same type signature as the real function. But something in here, either in mypy or in autospec or Mock, is causing that to be missed. Oh well.

auspicacious (approved):
Thank you!

tests/test_mock.py
master-of-ppap: Ohh, yeah. I see what you mean there. That is quite strange.

PR #85, opened 2024-12-02
Implement rounding for mpc320_steps using dimension

### LINKED ISSUE #83: Update mpc320_step type to output integers
In https://github.com/moonshot-nagayama-pj/PnPQ/pull/79 , we introduced a new `mpc320_velocity` Pint type that correctly translates floating-point, unbounded `angle / time` values into the integer percentage values that the device actually understands, validating that they are between 10 and 100.

We can do the same for the `mpc320_step` type, so that we no longer have to remember to call `round()` every time we do a conversion. We should do this before implementing other device drivers, so that we have a good pattern to follow.

## Acceptance criteria
* `mpc320_step` values are always integers, rounded from the exact floating-point conversion
* Unit tests are updated to remove `.approx`
* Unit tests are updated to explicitly assert that `mpc320_step` values produced by conversion are of type `int`. This assert is also added to the `_velocity` tests if it is missing.
* `mpc320_step` values are NOT limited to between 0 and 170 degrees. We will need to decide what the appropriate layer to implement this limit is later -- there may be times when it is convenient or useful to allow larger values.

TomShen1234:
Closes #83

I found that probably the simplest way to implement rounding for our unit is to subclass the existing `Quantity` class, and implement the behavior there. After that, we'll use a custom `UnitRegistry` that is set to use the new quantity class. This should not break existing behaviors. 

One thing that may need to be discussed is that there's a dictionary that will have to be updated manually to tell the subclass what units to round. 

Other than that, I updated the tests which converts degrees to `mpc320_step` to remove the `.approx()`, and also added an extra one to ensure that the rounding actually took effect. 

I didn't set a bound for the unit (yet), that will also be up for discussions. 

**EDIT**: See discussions below, a new strategy has been implemented to achieve the same effect.


TomShen1234:
I know I haven't updated the existing drivers to remove the `round()` yet. Until then, the new parts can be examined.

tests/test_pint_units.py
```diff
@@ -26,14 +26,15 @@ def test_mpc320_step_to_angle_conversion(
         (-170, -1370),
         (0, 0),
         (170, 1370),
+        (169, 1362),  # This will be able to test the actual rounding
     ],
 )
 def test_angle_to_mpc320_step_conversion(
     test_angle: float, expected_mpc320_step: float
```
master-of-ppap, line 33: It may be good to change the type of `expected_mpc320_step` to be `int` now

TomShen1234: Thanks

TomShen1234:
@auspicacious Do let me know if this approach is good enough or not. 
Apart from rounding, having a custom registry could also allow us to implement some more application-specific custom behaviors in the future.

src/pnpq/units.py
```diff
@@ -1,10 +1,51 @@
...
+            | None
+        ) = None,
+        *contexts: Any,
+        **ctx_kwargs: Any,
+    ) -> PlainQuantity[Any]:
+        """
+        Override implementation to implement rounding behavior.
+        """
+        value = super().to(other, *contexts, **ctx_kwargs)
+        if other in self.rounded_units:
```
auspicacious, line 36: I noticed that you're including both the singular and plural forms in the list, `["mpc320_step", "mpc320_steps"]`.

Does `super().to()` normalize the form in some way? If we stop using `other` on this line and instead use `value.units` or even `value.is_compatible_with`, can we avoid having to have both the singular and plural forms?

src/pnpq/units.py
```diff
@@ -1,10 +1,51 @@
...
+            return cast(
+                Quantity, PnPQCustomQuantity(round(value.magnitude), value.units)
+            )
+        return cast(Quantity, value)
+
+
+# Custom registry
+class PnPQCustomUnitRegistry(pint.UnitRegistry):
+    Quantity: TypeAlias = PnPQCustomQuantity
```
auspicacious, line 45: I'm not sure of the significance, but the pattern being employed here is a bit different than the one recommended in the official documentation:

https://pint.readthedocs.io/en/stable/advanced/custom-registry-class.html

tests/test_pint_units.py
```diff
@@ -26,14 +26,15 @@ def test_mpc320_step_to_angle_conversion(
...
     ],
 )
 def test_angle_to_mpc320_step_conversion(
-    test_angle: float, expected_mpc320_step: float
+    test_angle: float, expected_mpc320_step: int
 ) -> None:
 
     mpc320_step = (test_angle * pnpq_ureg.degree).to("mpc320_step").magnitude
-    assert mpc320_step == pytest.approx(expected_mpc320_step)
+    assert mpc320_step == expected_mpc320_step
```
auspicacious, line 37: You also need to explicitly assert that `mpc320_step` is an `int`.

auspicacious (commented):
@TomShen1234 I'm going to need some more details as to why this is better than defining contexts, which are designed to allow conversion between incompatible dimensions.

If I understand correctly, the goal here is to stop us from repeating ourselves when we have to implement this same rounding logic for other devices and other types. But we could easily do that using contexts as well. We also have more than one type of rounding behavior: velocity is different and requires more checks. Implementing all of this in `to` will make it quite complex.

Overriding a core part of the Pint library should only be done if we're confident there's no way to do it within the APIs that Pint already provides; changing `to`'s behavior could cause unexpected problems, especially when [Pint is doing some non-standard things in these classes](https://stackoverflow.com/questions/57429429/subclassing-q-in-pint-why-does-the-super-call-not-take-arguments). It also slows down all conversion operations since we now have to do a string comparison on every call to `to`, even when our custom units are not involved.

TomShen1234:
I fixed the small errors.

Regarding the actual way of implementation, are you sure there's an easy way to use the context to solve this? Seems like pint is insisting on using the default conversion method instead of the custom context transformation. 

```python
# Test codes
pnpq_ureg.define("mpc320_test_step = (170 / 1370) degree")
# Transformation function
def round_custom_units(ureg, value, **kwargs):
    print("Custom rounding called")
    return round(value)

# Create and register the context
custom_context = pint.Context("rounding_context")
custom_context.add_transformation("degree", "mpc320_test_step", round_custom_units)
pnpq_ureg.add_context(custom_context)
pnpq_ureg.enable_contexts("rounding_context")
```

```python3
test = (169 * pnpq_ureg.degree)
print(test)

test_mpc320_steps = test.to("mpc320_test_step")
print(test_mpc320_steps)
```

Seems like `round_custom_units` isn't being called at all, since there's no print message.

auspicacious:
@TomShen1234 I haven't tested it yet, but I think the trick here is to do the same as we've done for `mpc320_velocity` and make it dimensionless:

```
pnpq_ureg.define("mpc320_step = []")
```

then it should work. We then have to explicitly define transformations to and from `mpc320_step`.

If this makes your test code work, we should probably refactor the existing `mpc320_proportional_velocity` context to instead be called `apt_device_units` or something and use it to contain all conversions for APT devices.

auspicacious:
@TomShen1234 Alternatively, you could try [unit redefinition](https://pint.readthedocs.io/en/stable/user/contexts.html#using-contexts-for-unit-redefinition), which would preserve the dimension. I note that rounding is explicitly mentioned as a use case here.

TomShen1234:
@auspicacious I also thought this would work:

```python3
pnpq_ureg.define("mpc320_test_step = []")
...
def round_custom_units(ureg: pint.UnitRegistry, value: PlainQuantity[Quantity], **kwargs: Any) -> PlainQuantity[Any]:
    print("Custom rounding called")
    return Quantity(round(value.magnitude * 1370 / 190), value.units)
...
```

Now `pint` is just giving strange behaviors:

```
169 degree
2.949606435870417 mpc320_test_step
```

I think it's now trying to convert it to radians, still not calling my custom context function. 

Maybe I'll give the unit redefinition a shot.

TomShen1234:
Switched to use context transformations instead.

src/pnpq/units.py
```diff
@@ -1,21 +1,41 @@
...
+    return Quantity(round(value.magnitude * 1370 / 170), ureg.mpc320_step)
+
+
+def mpc320_steps_to_degree(
+    ureg: pint.UnitRegistry, value: PlainQuantity[Quantity], **_: Any
+) -> PlainQuantity[Any]:
+    return Quantity(value.magnitude * 170 / 1370, ureg.degree)
+
+
+# Create and register the context
```
auspicacious, line 29:
```suggestion
```

src/pnpq/units.py
```diff
@@ -34,7 +54,7 @@ def to_mpc320_velocity(
 
     converted_velocity = (velocity_in_degrees / mpc320_max_velocity) * 100
     rounded_velocity: Quantity = (
-        int(round(converted_velocity).magnitude) * ureg.dimensionless
+        int(round(converted_velocity).magnitude) * ureg.mpc320_velocity
```
auspicacious, line 57: This line is actually really funky. I think this can be simply written as
```suggestion
        round(converted_velocity.magnitude) * ureg.mpc320_velocity
```

src/pnpq/units.py
```diff
@@ -45,18 +65,66 @@ def to_mpc320_velocity(
     return rounded_velocity
 
 
-context.add_transformation(
+def mpc320_velocity_to_speed(
```
auspicacious, line 68: [Velocity is what pint calls it by default](https://github.com/hgrecco/pint/blob/master/pint/default_en.txt#L242), let's not confuse things by calling it speed as well in the same line.

```suggestion
def mpc320_velocity_to_velocity(
```

is probably fine, or

```suggestion
def mpc320_velocity_to_pint_velocity(
```

src/pnpq/units.py
```diff
@@ -45,18 +65,66 @@ def to_mpc320_velocity(
...
     "mpc320_velocity",
     "degree / second",
-    lambda ureg, value, **kwargs: (value * mpc320_max_velocity)
-    / 100,  # Convert value from percent
+    mpc320_velocity_to_speed,
+)
+
+
+# Add transformations between mpc320_velocity and mpc320_step
+def mpc320_velocity_to_mpc320_step_speed(
```
auspicacious, line 94:
```suggestion
def mpc320_velocity_to_mpc320_step_velocity(
```

src/pnpq/units.py
```diff
@@ -45,18 +65,66 @@ def to_mpc320_velocity(
...
+) -> PlainQuantity[Quantity]:
+    """
+    Converts an mpc320 velocity percentage to a velocity in degrees per second.
+    """
+    degrees_per_second = value.to("degree / second").magnitude
+    steps_per_second = degrees_per_second / 170 * 1370
+    return Quantity(steps_per_second, cast(Unit, ureg("mpc320_step / second")))
+
+
+def mpc320_step_speed_to_mpc320_velocity(
```
auspicacious, line 105:
```suggestion
def mpc320_step_velocity_to_mpc320_velocity(
```

test.py
```diff
@@ -0,0 +1,34 @@
+from pnpq.units import pnpq_ureg
```
auspicacious, line 1: Delete this file.

auspicacious (approved):

PR #88, opened 2024-12-06
Add py.typed, run uv lock --upgrade, add cache pruning to CI

auspicacious:
Two of the changes in this PR (uv.lock and cache pruning) are based on information I learned from reading the `uv` documentation.

`uv lock --upgrade` updates dependencies to the highest compatible version: something we should do on a regular basis.


master-of-ppap:
Sounds good. Thank you for including me in the review. I am not too sure what I should write here but, reading the typing documentation and a little about uv, it seems like a good idea.

auspicacious:
@master-of-ppap Thanks, could you approve this PR?

master-of-ppap (approved):

PR #90, opened 2024-12-10
Adds status update messages for waveplates

### LINKED ISSUE #89: Implement new messages for Thorlabs K10CR1 waveplate
The Thorlabs K10CR1 communicates with the computer using a different message than the MPC320. MPC320 uses `MGMSG_MOT_GET_USTATUSUPDATE`, while the K10CR1 uses `MGMSG_MOT_GET_STATUSUPDATE`. The goal is to rename the MPC320 messages to be ones that start with U, and then implement the `MGMSG_MOT_GET_STATUSUPDATE` and `MGMSG_MOT_REQ_STATUSUPDATE` messages. 

## Acceptance Criteria
* Rename `StatusBits` and `Status` to `UStatusBits` and `UStatus` in `protocol.py`.
* Add a new `StatusBits` and `Status` class to match the reply bitmask data of `MGMSG_MOT_GET_STATUSUPDATE` (p.123 in the APT communications protocol).
* Add `MGMSG_MOT_GET_STATUSUPDATE` and `MGMSG_MOT_REQ_STATUSUPDATE` messages class in `protocol.py` appropriately.
* Add appropriate unit tests.
  * Use the data from the `sample.log` file below to make the tests.
  * Or the sample data from the manual. But it's all zeroes.

## References
* [sample.log](https://github.com/user-attachments/files/18035252/sample.log)
* [apt communications protocol](https://www.thorlabs.com/Software/Motion%20Control/APT_Communications_Protocol.pdf)

TomShen1234:
Closes #89 

The following 2 messages were added:

- `MGMSG_MOT_GET_STATUSUPDATE`
- `MGMSG_MOT_REQ_STATUSUPDATE`

Seems like there's quite a lot of repetitions within the protocol files now, it may or may not be nice to abstract out some parts which are reused multiple times. 

Also, the definition for the status bit isn't too well written, there's no name (unlike its `USTATUSUPDATE` counterpart). Also, it ends with this interesting note: 'This is not full list of all the bits but the remaining bits reflect information about the state of
the hardware that in most cases does not affect motion.'

Please review the accuracy of the messages created.


src/pnpq/apt/protocol.py
```diff
@@ -225,6 +228,67 @@ class Status:
...
+
+
+@enum.unique
+class StatusBits(IntFlag, boundary=STRICT):
+    """Bitmask used in MGMSG_MOT_GET_USTATUSUPDATE to indicate motor
+    conditions. In the official documentation, all of these names have
+    P_MOT_SB prepended to them.
+
+    All spelling errors (e.g. "INITILIZING") are from the official
+    documentation.
```
auspicacious, line 254:
```suggestion
    conditions. In the official documentation, these values are not given clear, parseable names. However, nearly all of them correspond to values defined for the slightly different ``MGMSG_MOT_GET_USTATUSUPDATE`` command that do have clear names; we re-use those names here.
```

src/pnpq/apt/protocol.py
```diff
@@ -475,6 +539,70 @@ def to_bytes(self) -> bytes:
         )
 
 
+@dataclass(frozen=True, kw_only=True)
+class AptMessageWithDataEncCount(AptMessageWithData):
+    data_length: ClassVar[int] = 14
+
+    # There are some more bytes at the end according to the documentation:
+    # (WORD - channel 2 identifier, LONG, LONG, LONG)
+    # However, it says that these are for future use and should be ignored.
```
auspicacious, line 548: This comment is a bit confusing. It would be better to explicitly call out that the documentation says the message is 34 bytes long, but the actual message sent by the specific waveplate model we know about is 20 bytes.

"Ignoring" bytes would mean to me that we read 34 bytes and then don't process the extra 14 bytes. But that's not the case. The message is in fact only 20 bytes long. This is a critical difference, since if we failed to read 14 bytes that were actually there, we would not be able to correctly process the next message on the wire either.

tests/apt/test_protocol.py
```diff
@@ -290,6 +293,49 @@ def test_AptMessage_MGMSG_MOT_REQ_POSCOUNTER_to_bytes() -> None:
     assert msg.to_bytes() == b"\x11\x04\x01\x00\x50\x01"
 
 
+def test_AptMessage_MGMSG_MOT_GET_STATUSUPDATE_from_bytes() -> None:
+    msg = AptMessage_MGMSG_MOT_GET_STATUSUPDATE.from_bytes(
+        bytes.fromhex("8104 0e00 81 22 01000100 000000000000 07000000")
```
auspicacious, line 298: The formatting here doesn't make it easy to understand which parts correspond to which values.
```suggestion
        bytes.fromhex("8104 0e00 81 22 0100 01000000 00000000 07000000")
```

Was this pulled out of the logs? If not, do we have any actual values that we can test with?

auspicacious (commented):
I haven't looked at the actual data that we logged from the device, but I think this looks mostly correct. I think I spotted one bug in the code; the rest of the comments are quality-related.

src/pnpq/apt/protocol.py
TomShen1234: That's actually exactly what the documentation says (although the documentation itself is somewhat poor)

tests/apt/test_protocol.py
TomShen1234: The `enc_count` part is taken from the documentation example, the rest are the same as the test for `USTATUSUPDATE` message. 
"Enc Counts: 00, 00, 00, 00: Only used with encoded stages"

master-of-ppap:
Thanks for the code. It looks very good to my eyes. Beauty is in the eye of the beholder as they say.

However, attempting to run the tests with the byte data from `sample.log` in the issue causes an error:

sample.log:
```json
{
  "message": "AptMessageForStreamParsing(message_id=1153, data_length=14)",
  "bytes": "b'\\x81\\x04\\x0e\\x00\\x81P\\x01\\x00V\\xd5\\xfb\\xff\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x80'",
  "event": "rx_message_unknown",
  "level": "debug",
  "timestamp": "2024-12-06T08:25:08.465419Z"
}
```

Updated code: (test_protocol.py:296)
```py
def test_AptMessage_MGMSG_MOT_GET_STATUSUPDATE_from_bytes() -> None:
    msg = AptMessage_MGMSG_MOT_GET_STATUSUPDATE.from_bytes(
        # bytes.fromhex("8104 0e00 81 22 01000100 000000000000 07000000")
        b"\x81\x04\x0e\x00\x81P\x01\x00V\xd5\xfb\xff\x00\x00\x00\x00\x00\x00\x00\x80"
    )
```

Seems to throw an error upon running pytest. 
```py
======================================= short test summary info ========================================
FAILED tests/apt/test_protocol.py::test_AptMessage_MGMSG_MOT_GET_STATUSUPDATE_from_bytes - ValueError: <flag 'StatusBits'> invalid value 2147483648
```

Not too sure what the problem is, but it may be worth looking into.

src/pnpq/apt/protocol.py
auspicacious: I know that documentation uses the word "ignore." I'm explaining why "ignore" doesn't necessarily mean "don't read."

TomShen1234:
@master-of-ppap 
After some investigation, we realized that the problem lies in `boundary=STRICT` in the `IntFlag` documentation. We have decided that the current plan would be to remove that, and just have unknown value stay as-is.

auspicacious (approved):

PR #94, opened 2025-01-13
Split MOVE_COMPLETED into two separate dataclasses for 6-byte and 20-byte messages.

master-of-ppap:
I have been pulling my hair out on trying to find an elegant way of separating the dataclasses, and I think I found an alright solution.

Additions:
* Added two new classes `AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES` and `AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES` that inherit from a main `AptMessage_MGMSG_MOT_MOVE_COMPLETED` class. These reflect the two different possibilities of `MOVE_COMPLETED`  messages for the Thorlabs devices. One type contains only 6 header bytes, while the other type contains 6 bytes with an additional 14 data bytes. This can be seen on p.125 of the APT Communications Protocol.
* Added their respective unit tests. 
* Updated the existing MPC320 driver to use this new dataclass.

Considerations:
* As far as we know, the MPC320 prefers the 6 byte message, while the K10CR1 prefers the 20 byte message. However, we are still unsure of what future devices will prefer, so I have left the names ambiguous for that reason. But, the current names are kind of dumb, so if you have any better suggestions I will be grateful.

Resources:
* [APT Communications Protocol](https://www.thorlabs.com/Software/Motion%20Control/APT_Communications_Protocol.pdf).

### DIFF
--- a/src/pnpq/apt/protocol.py
+++ b/src/pnpq/apt/protocol.py
@@ -991,27 +991,41 @@ class AptMessage_MGMSG_MOT_MOVE_COMPLETED(AptMessage):
     One for the full 20 byte message, and another for the 6 byte message.
     """
 
+    @classmethod
+    def from_bytes(cls, raw: bytes) -> "AptMessage_MGMSG_MOT_MOVE_COMPLETED":
+        length = len(raw)
+        if length == 6:
+            return AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES.from_bytes(raw)
+        if length == 20:
+            return AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES.from_bytes(raw)
+        raise ValueError(
+            f"Expected data packet length 6 or 20, but received {length} instead. Full raw data was {raw!r}"
+        )
+        pass
+
 
 @dataclass(frozen=True, kw_only=True)
 class AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES(
-    AptMessage_MGMSG_MOT_MOVE_COMPLETED, AptMessageHeaderOnlyChanIdent
+    AptMessageHeaderOnlyChanIdent, AptMessage_MGMSG_MOT_MOVE_COMPLETED
 ):
     """
     For the MPC320, no data packet follows the main move completed message, so this message is used.
     """
 
     message_id: ClassVar[AptMessageId] = AptMessageId.MGMSG_MOT_MOVE_COMPLETED
+    data_length: ClassVar[int] = 0
 
 
 @dataclass(frozen=True, kw_only=True)
 class AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES(
-    AptMessage_MGMSG_MOT_MOVE_COMPLETED, AptMessageWithDataMotorStatus
+    AptMessageWithDataMotorStatus, AptMessage_MGMSG_MOT_MOVE_COMPLETED
 ):
     """
     For the K10CR1, a full USTATUS data packet follows the main move completed message, so this message is used.
     """
 
     message_id: ClassVar[AptMessageId] = AptMessageId.MGMSG_MOT_MOVE_COMPLETED
+    data_length: ClassVar[int] = 14
 
 
 @dataclass(frozen=True, kw_only=True)
--- a/src/pnpq/apt/protocol.py
+++ b/src/pnpq/apt/protocol.py
@@ -991,6 +991,8 @@ class AptMessage_MGMSG_MOT_MOVE_COMPLETED(AptMessage):
     One for the full 20 byte message, and another for the 6 byte message.
     """
 
+    data_length: ClassVar[int]
+
     @classmethod
     def from_bytes(cls, raw: bytes) -> "AptMessage_MGMSG_MOT_MOVE_COMPLETED":
         length = len(raw)
@@ -1001,7 +1003,6 @@ class AptMessage_MGMSG_MOT_MOVE_COMPLETED(AptMessage):
         raise ValueError(
             f"Expected data packet length 6 or 20, but received {length} instead. Full raw data was {raw!r}"
         )
-        pass
 
 
 @dataclass(frozen=True, kw_only=True)
@@ -1015,6 +1016,10 @@ class AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES(
     message_id: ClassVar[AptMessageId] = AptMessageId.MGMSG_MOT_MOVE_COMPLETED
     data_length: ClassVar[int] = 0
 
+    @classmethod
+    def from_bytes(cls, raw: bytes) -> Self:
+        return super().from_bytes(raw)
+
 
 @dataclass(frozen=True, kw_only=True)
 class AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES(
@@ -1026,6 +1031,10 @@ class AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES(
 
     message_id: ClassVar[AptMessageId] = AptMessageId.MGMSG_MOT_MOVE_COMPLETED
     data_length: ClassVar[int] = 14
+
+    @classmethod
+    def from_bytes(cls, raw: bytes) -> Self:
+        return super().from_bytes(raw)
 
 
 @dataclass(frozen=True, kw_only=True)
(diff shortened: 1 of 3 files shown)


tests/apt/test_protocol.py
```diff
@@ -440,15 +441,45 @@ def test_AptMessage_MGMSG_MOT_MOVE_COMPLETED_from_bytes() -> None:
...
+
+def test_AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES_to_bytes() -> None:
+    msg = AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES(
+        chan_ident=ChanIdent.CHANNEL_1,
+        destination=Address.HOST_CONTROLLER,
+        source=Address.GENERIC_USB,
+        position=0x01A0AA68,
+        velocity=0x00,
+        motor_current=pnpq_ureg.milliamp * 0,
+        status=UStatus(INMOTIONCCW=True, INMOTIONCW=True, ENABLED=True),
```
auspicacious, line 476: In motion both clockwise and counterclockwise at the same time... well, it's not worth worrying about now if it parses.

tests/apt/test_protocol.py
```diff
@@ -440,15 +441,45 @@ def test_AptMessage_MGMSG_MOT_MOVE_COMPLETED_from_bytes() -> None:
...
+
+
+def test_AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES_to_bytes() -> None:
+    msg = AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES(
+        chan_ident=ChanIdent.CHANNEL_1,
+        destination=Address.HOST_CONTROLLER,
+        source=Address.GENERIC_USB,
+        position=0x01A0AA68,
+        velocity=0x00,
+        motor_current=pnpq_ureg.milliamp * 0,
```
auspicacious, line 475: The style you'll see in [the pint manual](https://pint.readthedocs.io/en/stable/getting/overview.html) writes this in the same order we would use in English or Japanese:

```suggestion
        motor_current=0 * pnpq_ureg.milliamp,
```

tests/apt/test_protocol.py
```diff
@@ -440,15 +441,45 @@ def test_AptMessage_MGMSG_MOT_MOVE_COMPLETED_from_bytes() -> None:
...
+    assert msg.motor_current == 0x00
+    assert msg.status == UStatus(INMOTIONCCW=True, INMOTIONCW=True, ENABLED=True)
+
+
+def test_AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES_to_bytes() -> None:
+    msg = AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES(
+        chan_ident=ChanIdent.CHANNEL_1,
+        destination=Address.HOST_CONTROLLER,
+        source=Address.GENERIC_USB,
+        position=0x01A0AA68,
```
auspicacious, line 473: I thought that this looked big, but apparently this is about halfway to the maximum number of steps.

https://www.thorlabs.com/newgrouppage9.cfm?objectgroup_id=8750 ("specs" tab on the right)

auspicacious (commented):
The naming (`6_BYTES`, `20_bytes`) is a good idea. This looks OK so far, but it's not complete until the `from_bytes` method is implemented.

src/pnpq/apt/protocol.py
```diff
@@ -984,10 +984,34 @@ def to_bytes(self) -> bytes:
...
-    followed by a full USTATUS data packet. In reality, for the
-    MPC320, no data packet follows."""
+class AptMessage_MGMSG_MOT_MOVE_COMPLETED(AptMessage):
+    """
+    Note that the APT documentation indicates that this should be
+    followed by a full USTATUS data packet. So, two separate methods will be defined.
+    One for the full 20 byte message, and another for the 6 byte message.
+    """
+
+    pass  # pylint: disable=W0107
```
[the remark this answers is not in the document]
master-of-ppap: Should we define a new `from_bytes` method here in this `AptMessage_MGMSG_MOT_MOVE_COMPLETED` superclass? The two subclasses inherit from other classes that automatically provide the `from_bytes` and `to_bytes` method, so I thought it wouldn't be necessary here. Should I redefine `from_bytes` as an `@abstractmethod`?

tests/apt/test_protocol.py
master-of-ppap: Yeah... I'm not too sure what that's supposed to mean, but that's what the data came out as.

tests/apt/test_protocol.py
master-of-ppap: Oh yes, this device is extremely precise it seems. That price tag makes me realize I must take extra care of it...

master-of-ppap:
Finished making changes. Plz see my question here.
https://github.com/moonshot-nagayama-pj/PnPQ/pull/94#discussion_r1914216297

Thank you.

src/pnpq/apt/protocol.py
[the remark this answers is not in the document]
auspicacious: @master-of-ppap Take a look at the code in `connection.py` that instantiates the class (`rx_dispatch` around line 189) and I think it will be clear why the parent class needs a `from_bytes` method.

It doesn't need to be an `@abstractmethod`.

Basically, if you call `from_bytes` on `AptMessage_MGMSG_MOT_MOVE_COMPLETED`, it should give you either a `AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES` or a `AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES`.

If you call `from_bytes` on `AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES`, it should only give you a `AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES`.

master-of-ppap:
> @master-of-ppap Take a look at the code in `connection.py` that instantiates the class (`rx_dispatch` around line 189) and I think it will be clear why the parent class needs a `from_bytes` method.
> 
> It doesn't need to be an `@abstractmethod`.
> 
> Basically, if you call `from_bytes` on `AptMessage_MGMSG_MOT_MOVE_COMPLETED`, it should give you either a `AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES` or a `AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES` depending on how many bytes are passed in.
> 
> If you call `from_bytes` on `AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES`, it should only give you a `AptMessage_MGMSG_MOT_MOVE_COMPLETED_6_BYTES`.

Yes. I see what you mean. That's actually a really good idea. 

I've tried implementing that in this recent commit, along with some test code. The code is fine, and the tests work. But there seems to be many `mypy` typing issues, and I have no idea how I'm supposed to fix them...

Could you provide some assistance? Haha.
Thank you.

auspicacious:
@master-of-ppap I'll check out your code and take a look. If I can fix it easily I'll commit and push it to the PR. I've run into annoyances before when trying to use polymorphism in this way.

auspicacious (approved):
Fixed it. As long as we don't change the order that the inherited classes are listed in, this should work. Feel free to merge it.

This feels like a mypy bug to me, and indeed [a somewhat similar issue](https://github.com/python/mypy/issues/18458) was reported less than a day ago. That issue gave me the inspiration to try solving this by overriding `to_bytes` in the subclasses.

Regardless of whether it's a bug or not, this is actually a good illustration of the limits of using multiple inheritance to share functionality, and also the limits of polymorphism.

The type checker can't infer whether the 20_BYTE or 6_BYTE subclass is instantiated when `to_bytes` is called, which means we have to manually [type narrow](https://mypy.readthedocs.io/en/stable/type_narrowing.html) the result afterward to use any of the properties that are only present on the 20-byte version. In general, this is a code smell. Ideally, subclasses conform to the same public interface as their parent class and we do not have to cast them to the concrete subclass to use them.

That being said, I can't think of a better way to process a stream of heterogeneous messages like this using object-oriented design.

master-of-ppap:
Oh my gosh. I don't want to know how long it took for you to find and debug that. That seems like a painful task. I apologize for bringing this upon you.

Anyway, I really appreciate you solving it for me. Python type linting is very painful.
Thank you.

Issue #28, opened 2024-07-17
Refactor PnPQ APIs to be context managers


auspicacious:
Most devices managed by PnPQ require some sort of connection to be opened before commands are sent. This is perfect for the [context manager pattern](https://realpython.com/python-with-statement/) in Python, e.g. the `with` statement.

## Prerequisites

* PnPQ drivers should already be refactored to inherit from a common interface before starting this work. Implementing this pattern without a common interface will only cause more churn. See https://github.com/moonshot-nagayama-pj/PnPQ/issues/29
* We should validate that code in testbed-manager, testbed-dashboard, etc. will work well with this pattern.

## Acceptance criteria

* PnPQ drivers now work with the `with` statement as appropriate for each device
* Unit tests updated or written to test this pattern
* Existing testbed code updated to use this pattern

master-of-ppap:
Do we still need this? It's kind of old, but I need something to do, haha.

auspicacious:
@master-of-ppap If you want to learn how context managers work, sure, go ahead and see if you can implement something for the MPC.

PR #99, opened 2025-01-31
Implement close() method for AptConnection

master-of-ppap:
One step towards closing #28.

In order to implement context managers in PnPQ, we must be able to close the AptConnection at will. This is what I have tried to implement.

## Additions
* `close()` method for `AptConnection` that stops all polling threads and then flushes and closes the serial connection.
* Implemented new `test_connection` method in MPC320 driver to test this function.
  *  Tested new `AptConnection.close()` against MPC320 device using this driver test function, and it passes! Let's go!
* Updated python version to 3.13.1 to be able to use `Queue.shutdown()`
* Changed queue implementation in `AptConnection` to make use of `Queue` instead of `SimpleQueue` to use the `Queue.shutdown()` method.


master-of-ppap:
Finished making changes. Thank you to all of my fans.

master-of-ppap:
Haha. Just some classic humor. This is actually the last commit. Got you.

src/pnpq/apt/connection.py
```diff
@@ -155,44 +176,52 @@ def __post_init__(self) -> None:
...
-    # manager for this object.
-    #
-    # The tricky part is managing the threads; they will all need to
-    # be paused before closing the serial connection... and then how
-    # do we clean up the child threads if this object gets cleaned up?
+    def close(self) -> None:
 
-    # def __enter__(self) -> None:
-    #     self.open()
+        self.send_message_no_reply(
```
auspicacious, line 181: This should probably be sent immediately using the unordered send function.

src/pnpq/apt/connection.py
```diff
@@ -155,44 +176,52 @@ def __post_init__(self) -> None:
...
+            AptMessage_MGMSG_HW_STOP_UPDATEMSGS(
+                destination=Address.GENERIC_USB,
+                source=Address.HOST_CONTROLLER,
+            )
+        )
 
-    # def __exit__(self, exc_type, exc_value, exc_tb) -> None:
-    #     self.close()
+        time.sleep(1)
```
auspicacious, line 188: Let's try it without the `sleep`.

src/pnpq/apt/connection.py
```diff
@@ -155,44 +176,52 @@ def __post_init__(self) -> None:
...
         with self.rx_dispatcher_thread_lock:
-            while True:
+            while not self.stop_event.is_set():
+
                 partial_message: None | AptMessageForStreamParsing = None
                 full_message: Optional[AptMessage] = None
                 try:
                     message_bytes = self.connection.read(6)
+                # Serial bus not connected error
+                except Exception:  # pylint: disable=W0718
```
auspicacious, line 211: Probably a debug-level log of this exception is a good idea. You can use `exc_info` to log the exception.

src/pnpq/devices/polarization_controller_thorlabs_mpc.py
```diff
@@ -93,8 +96,10 @@ def tx_poll(self) -> None:
                 # update by setting the
                 # tx_ordered_sender_awaiting_reply event.
                 if self.connection.tx_ordered_sender_awaiting_reply.is_set():
+
                     time.sleep(0.2)
                 else:
+
```
auspicacious, line 102: Drop some of these empty lines in if/else statements.

auspicacious (changes requested):

auspicacious (approved):

PR #101, opened 2025-02-07
Finished implementation of context manager

### LINKED ISSUE #28: Refactor PnPQ APIs to be context managers

master-of-ppap:
Closes #28.

## Additions
* Adds `__enter__` and `__exit__` methods for the `AptConnection` class, so we can use them as context managers using the `with` keyword.
* Updated the existing unit tests to fit this pattern. 
* Tested with the real MPC320 device, and no errors are thrown for all new hardware tests.


auspicacious (approved):
Thanks, I think this is good enough for now. I've created https://github.com/moonshot-nagayama-pj/PnPQ/issues/107 to cover any other work related to clean connection shutdown.

PR #108, opened 2025-02-21
Fix bug that made non-degree type angles be incorrectly converted to mpc320_step 

master-of-ppap:
While messing around with testbed-node last week with Tom, we encountered a strange error in which radians converted to type `mpc320_step` visually resulted in the wrong angle on the device. I was looking through the `units.py` code and found the culprit of this issue.

### Issue

When converting specific angles into `mpc320_step`, we perform the conversion based on the assumption that the input angle is in degrees. However, this may not necessarily be the case. 

### Fix

Therefore, an easy fix is to convert the inputted angle into degrees first, then perform the conversion calculation to `mpc320_step`.

### Additions
I have implemented this quick fix and also updated the unit tests to accordingly test for conversions from non-degree type angles (such as radians).


auspicacious (approved):
Glad that you caught this.

PR #109, opened 2025-02-22
Initial ABC Implementation for MPC

### LINKED ISSUE #105: Abstract Base Class for MPC
It would be nice to create an abstract base class (ABC) for each devices so that our API stays relevantly consistent, even if we decided to add similar devices in the future. This will start with the MPC, after we figure out a pattern, it can be extended to the other devices. 

After this is done, we can use it to create a stub device for the MPC. 

## Acceptance Criteria

- An ABC is created for the MPC with the appropriate function definitions
- The existing driver class should become a subclass of the ABC

TomShen1234:
Close #105 

I made a new file that contains the ABC for the MPC class, and the existing class is made to subclass the ABC. 

Existing documentation and shared properties (except `connection`, comments below) are removed from the actual class implementation to use the one from the ABC. Also, one helper function (`get_status_all`) has also been moved to the ABC class. 

It seems like removing the re-declaration of `connection` from the actual implementation dataclass will cause the initializer synthesis to fail, so I kept it there. Maybe there's a better approach to this (or the entire handling of properties)?

The `PolarizationControllerParams` definition has also been moved to the base.


src/pnpq/devices/polarization_controller_base.py
```diff
@@ -0,0 +1,92 @@
...
+        pass
+
+    @abstractmethod
+    def set_params(
+        self,
+        velocity: None | Quantity = None,
+        home_position: None | Quantity = None,
+        jog_step_1: None | Quantity = None,
+        jog_step_2: None | Quantity = None,
+        jog_step_3: None | Quantity = None,
```
auspicacious, line 90: I note that this is a little less than ideal since not all devices have three paddles, but probably not a problem in practice for us. (No changes necessary here).

src/pnpq/devices/polarization_controller_base.py
```diff
@@ -0,0 +1,92 @@
...
+    # Setup channels for the device
+    available_channels: frozenset[ChanIdent] = frozenset([])
+
+    def get_status_all(self) -> tuple[AptMessage_MGMSG_MOT_GET_USTATUSUPDATE, ...]:
+        all_status = []
+        for channel in self.available_channels:
+            status = self.get_status(channel)
+            all_status.append(status)
+        return tuple(all_status)
+
```
auspicacious, line 42: The easiest way to avoid issues is probably to remove all implementation from the ABC and use it purely as an abstract base class.

Consider that the constructor may be different between implementations: our current real-device implementations need an `AptConnection`, but the stub device does not need an `AptConnection`. If we somehow had a third, non-APT device use this interface (unlikely, but maybe possible with refactoring), it would also not need an `AptConnection`.

```suggestion
```

src/pnpq/devices/polarization_controller_base.py
```diff
@@ -0,0 +1,92 @@
...
+    home_position: Quantity
+    #: Dimensionality must be [angle] or mpc320_step
+    jog_step_1: Quantity
+    #: Dimensionality must be [angle] or mpc320_step
+    jog_step_2: Quantity
+    #: Dimensionality must be [angle] or mpc320_step
+    jog_step_3: Quantity
+
+
+class PolarizationControllerBase(ABC):
```
auspicacious, line 28: About names and locations: it's probably better to keep all of the Thorlabs MPC-related code in the same file. The file will get huge, but it makes sense that all of this code should be part of the same Python module. (There are ways to split a module across multiple files, but let's not do that right now).

I'd suggest moving the contents of this file into `polarization_controller_thorlabs_mpc` and renaming the ABC to be more consistent with the other class names so that it's specific to Thorlabs. `AbstractPolarizationControllerThorlabsMPC` could work. (Arguably we should drop the "ThorlabsMPC" bit from all of the names since they're all in a module that makes that clear, but then we would have to refactor more code).

auspicacious (commented):
I think you've identified the main challenges in using an ABC, some suggested changes in inline comments. In the future we might want to put the stub implementation into this same module as well.

src/pnpq/devices/polarization_controller_base.py
TomShen1234: I see, what about the stub devices? Shall we use the same structure as before where it's a separate file, or should we also have it in the same file (since it'll probably be a ThorLabs MPC stub rather than a generic MPC stub)

auspicacious (approved):
Thanks, I think this will do nicely as a start.

I could understand arguments for putting the stubs into a separate directory at the same level as `devices`, so that we would have `pnpq/stubs` and `pnpq/devices` (and maybe `pnpq/interfaces` for the ABCs). I could also understand arguments for putting the stubs in this same file for now.

There's really no perfect answer. In Java where one file == one class the first directory structure I suggested is commonly used, but in Python we have much more flexibility in organizing modules.

PR #110, opened 2025-03-07
Add K10CR1 mock waveplate and refactor pint units code to use generator

### LINKED ISSUE #103: Implement mock driver tests for K10CR1 optical waveplate
In order to add unit tests for the K10CR1 optical waveplate, we should implement another mock that can read the messages sent by the drivers, and send the corresponding response, similar to as was done in #74. 

It should behave the same way as the already existing mock driver for the MPC320 in `PnPQ/tests/test_polarization_controller_thorlabs_mpc320.py.`

We should keep it simple, and support just the `move_absolute()` function of the K10CR1 for now.

## Acceptance Criteria
* A mock for `WaveplateThorlabsK10CR1` created.
* It should accept the `AptMessage_MGMSG_MOT_MOVE_ABSOLUTE` and match the correct `AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES` as would have been sent from the real device.
* Unit test should be created for the move_absolute function of the driver.
* Assert that `send_message_expect_reply` was called the expected number of times.

master-of-ppap:
Closes #103 

This should have been split into two different pull requests but I worked on them on the same branch, and I'm not sure if it's possible to split it now... So, I'll leave it like this for now. Please forgive me.

## Implementations
* Adds a mock waveplate for the K10CR1 that verifies the `move_absolute` method in the base driver listens for the appropriate message
* Since the `units.py` class was getting a bit repetitive, I tried to encapsulate the repeated code into a function `get_unit_transformation` which generates the appropriate transformation units. 
  * Also has flexibility to allow for complex transformations using a lambda function, as well as the ability to round the output or limit the output to a certain range.

### DIFF
--- a/src/pnpq/units.py
+++ b/src/pnpq/units.py
@@ -9,6 +9,7 @@ pnpq_ureg = pint.UnitRegistry()
 
 thorlabs_context = pint.Context("thorlabs_context")
 
+
 def get_unit_transformation(
     input_to_output: Callable[[float], float],
     input_unit: pint.Unit,
@@ -38,6 +39,7 @@ def get_unit_transformation(
         return converted_quantity
 
     return to_quantity
+
 
 # Custom unit definitions for devices
 pnpq_ureg.define("mpc320_step = [dimension_mpc320_step]")
@@ -119,8 +121,6 @@ thorlabs_context.add_transformation(
         input_to_output=lambda steps: steps / 100 * 400,
         input_unit=cast(Unit, pnpq_ureg("mpc320_velocity")),
         output_unit=cast(Unit, pnpq_ureg("degree / second")),
-        output_range=None,
-        output_rounded=False,
     ),
 )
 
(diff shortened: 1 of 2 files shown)


auspicacious (approved):
@master-of-ppap This code looks good, but I think that you forgot to commit the mock waveplate code.

I'm approving this, so you can merge it as it is, which will delete the remote branch.

Then you can commit the waveplate code and push the branch again, which will re-create the remote branch. You can make a second PR for the waveplate code.

(Just be careful not to delete your local work).

master-of-ppap:
@auspicacious Thanks for reviewing the code.

I may have some memory disease. It seems as though I have already created the K10CR1 four months ago...
https://github.com/moonshot-nagayama-pj/PnPQ/commit/3bb53a64ef15c89211df891528615aa9b5c71705#diff-b4c0c72fafdb86f3e3850820c34c20f5e095710fc9c52a691dfddded65759811

PR #117, opened 2025-04-11
Improved Documentation

### LINKED ISSUE #113: Add documentations to finished public interfaces
As the different drivers become more complete, it's should now be a good time to add documentations for them. We'll start with the MPC320 and our refactored Waveplates driver. 

This should also be a good time to figure out how our documentations should look like. It will probably follow the pattern of the `jog` function of MPC320. 

## Acceptance Criteria

- Documentations added to all public functions of the ABCs of the MPC and Waveplates
- The structures of the documentation should be updated slightly to make the website look nicer

TomShen1234:
Closes #113 

Added:
- Documentation to MPC320
- Fixed documentation for waveplate
- Made homepage of documentation better
- Added contributing.md

Suggestions are welcomed (especially on grammars and word choice). 

I think after this, a further task still needs to be added to complete the rest of the documentation (for the other classes).

### DIFF
--- a/sphinx/source/index.rst
+++ b/sphinx/source/index.rst
@@ -13,3 +13,22 @@ PnPQ
    api/modules
 
 PnPQ is a control library for hardware commonly used in quantum optical experiments (although it is probably useful for classical optical experiments as well!).
+
+Supported devices
+-----------------
+
+**ThorLabs**:
+
+- :py:mod:`Waveplates (kbd101) <pnpq.devices.waveplate_thorlabs_kb10crm>`
+- :py:mod:`Optical Delay Line (kbd101) <pnpq.devices.odl_thorlabs_kbd101>`
+- :py:mod:`Optical Switch (osw1310e) <pnpq.devices.switch_thorlabs_osw1310e>`
+- :py:mod:`Motorized Polarization Controller (mpc320) <pnpq.devices.polarization_controller_thorlabs_mpc>`
+
+**OzOptics**:
+
+- :py:mod:`Optical Delay Line (650ml) <pnpq.devices.odl_ozoptics_650ml>`
+
+About the project
+-----------------
+
+`GitHub <https://github.com/moonshot-nagayama-pj/PnPQ>`_
--- a/sphinx/source/index.rst
+++ b/sphinx/source/index.rst
@@ -17,14 +17,16 @@ PnPQ is a control library for hardware commonly used in quantum optical experime
 Supported devices
 -----------------
 
-**ThorLabs**:
+ThorLabs
+^^^^^^^^
 
 - :py:mod:`Waveplates (kbd101) <pnpq.devices.waveplate_thorlabs_kb10crm>`
 - :py:mod:`Optical Delay Line (kbd101) <pnpq.devices.odl_thorlabs_kbd101>`
 - :py:mod:`Optical Switch (osw1310e) <pnpq.devices.switch_thorlabs_osw1310e>`
 - :py:mod:`Motorized Polarization Controller (mpc320) <pnpq.devices.polarization_controller_thorlabs_mpc>`
 
-**OzOptics**:
+OzOptics
+^^^^^^^^
 
 - :py:mod:`Optical Delay Line (650ml) <pnpq.devices.odl_ozoptics_650ml>`
 
--- a/src/pnpq/devices/polarization_controller_thorlabs_mpc.py
+++ b/src/pnpq/devices/polarization_controller_thorlabs_mpc.py
@@ -46,21 +46,48 @@ class PolarizationControllerParams(TypedDict):
 class AbstractPolarizationControllerThorlabsMPC(ABC):
     @abstractmethod
     def get_status_all(self) -> tuple[AptMessage_MGMSG_MOT_GET_USTATUSUPDATE, ...]:
-        pass
+        """Fetch the latest status of all channels on the device.
+
+        :param chan_ident: The motor channel to fetch status for.
+        :return: A tuple of
+            :py:class:`AptMessage_MGMSG_MOT_GET_USTATUSUPDATE`,
+            one for each channel.
+
+        """
 
     @abstractmethod
     def get_status(
         self, chan_ident: ChanIdent
     ) -> AptMessage_MGMSG_MOT_GET_USTATUSUPDATE:
-        pass
+        """Fetch the latest status of the device.
+
+        :param chan_ident: The motor channel to fetch status for.
+        :return: The message returned by the device, in
+            :py:class:`AptMessage_MGMSG_MOT_GET_USTATUSUPDATE`
+
+        """
 
     @abstractmethod
     def home(self, chan_ident: ChanIdent) -> None:
-        pass
+        """Move the device to home position
+
+        The home position can be customized using the
+        :py:func:`set_params` function.
+
+        :param chan_ident: The motor channel to set to home.
+
+        """
 
     @abstractmethod
     def identify(self, chan_ident: ChanIdent) -> None:
-        pass
+        """Identifies the device represented by this instance
+        by flashing the LED light on the device.
+
+        On a stub device, this function does nothing.
+
+        :param chan_ident: The motor channel to identify.
+
+        """
 
     @abstractmethod
     def jog(self, chan_ident: ChanIdent, jog_direction: JogDirection) -> None:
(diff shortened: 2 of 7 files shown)


src/pnpq/devices/polarization_controller_thorlabs_mpc.py
```diff
@@ -78,15 +105,35 @@ def jog(self, chan_ident: ChanIdent, jog_direction: JogDirection) -> None:
...
+
+        :return: The set of parameters in a dictionary defined
+            in :py:class:`PolarizationControllerParams`.
+
+        """
 
     @abstractmethod
     def set_channel_enabled(self, chan_ident: ChanIdent, enabled: bool) -> None:
-        pass
+        """Enables or disables a channel specified.
```
auspicacious, line 129:
```suggestion
        """Enables or disables the specified motor channel. End users will not typically use this command. Instead, commands that require a channel to be enabled will automatically enable the channel before executing, and disable the channel when complete.
```

src/pnpq/devices/polarization_controller_thorlabs_mpc.py
```diff
@@ -78,15 +105,35 @@ def jog(self, chan_ident: ChanIdent, jog_direction: JogDirection) -> None:
...
+
+        :param chan_ident: The motor channel to move.
```
auspicacious, line 131:
```suggestion
        :param chan_ident: The motor channel to enable.
```

src/pnpq/devices/polarization_controller_thorlabs_mpc.py
```diff
@@ -78,15 +105,35 @@ def jog(self, chan_ident: ChanIdent, jog_direction: JogDirection) -> None:
...
+        :param enabled: A boolean value that specifies whether the
+            devices are enabled or disable. `True` will enable the
+            channel and vice versa.
```
auspicacious, line 134: I think this can be replaced with

```
Set to `True` to enable the channel, or `False` to disable.
```

src/pnpq/devices/polarization_controller_thorlabs_mpc.py
```diff
@@ -46,21 +46,48 @@ class PolarizationControllerParams(TypedDict):
 class AbstractPolarizationControllerThorlabsMPC(ABC):
     @abstractmethod
     def get_status_all(self) -> tuple[AptMessage_MGMSG_MOT_GET_USTATUSUPDATE, ...]:
-        pass
+        """Fetch the latest status of all channels on the device.
+
+        :param chan_ident: The motor channel to fetch status for.
```
auspicacious, line 51:
```suggestion
```

src/pnpq/devices/polarization_controller_thorlabs_mpc.py
```diff
@@ -97,7 +144,27 @@ def set_params(
         jog_step_2: None | Quantity = None,
         jog_step_3: None | Quantity = None,
     ) -> None:
-        pass
+        """Update the parameters of the device.
+
+        All parameters of this function are optional. Only fields
+        with values are updated on the device.
+
+        :param velocity: The velocity which the device will move.
```
master-of-ppap, line 152:
```suggestion
        :param velocity: The velocity which the device will move at.
```

master-of-ppap (commented):
LGTM (Looks Good To Me)!

src/pnpq/devices/polarization_controller_thorlabs_mpc.py
auspicacious: @master-of-ppap Ending a sentence with a preposition? This is the sort of English up with which I will not put.

> The velocity at which the device will move.

In all seriousness, being parsimonious with words is a good idea in API documentation.

> The rotational velocity. Applies to all channels. Unit must be convertible to ``mpc320_velocity``.

Also... now that I look at this closer @TomShen1234 : does this render correctly? reStructuredText requires two backticks for code samples but only one is used here.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup

master-of-ppap: Dammit. My review got reviewed. That's painful.
Yes, I agree. Parsimoniousness would be a good call here.

TomShen1234:
As for the rendering problems:

\`mpc320_step\` renders like this: _mpc320_step_
\`\`mpc320_step\`\` renders like this: `mpc320_step`

Which one do you prefer?

The first one is used by the template to show types of variables like:
chan_ident (_ChanIdent_) - ...

auspicacious:
@TomShen1234 In general, it's a much more common convention to use fixed-width text when referring to code, and I don't want to try to distinguish between different use cases like the Sphinx template does, so two backticks it is.

To expand on that a little, looking at the current Sphinx template (which I assume is similar to the one you're updating to), it looks like they're trying to visually distinguish between function arguments and types by using italics + fixed-width for arguments and italics for types. Fixed-width is used almost everywhere else.

Typesafe languages tend to inline the type information and keep it in fixed-width, cf. [docs.rs](https://docs.rs/humantime/latest/humantime/struct.Duration.html) and [Javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/math/BigDecimal.html).

I realize that this means a lot of updates, including to text that I myself wrote earlier.

Speaking of which, I think the entire `enabled` docstring can be made much shorter (it looks like this comment got missed): https://github.com/moonshot-nagayama-pj/PnPQ/pull/117/commits/caafb3f19754e583050ea46a35367ffe1e075336#r2039074200

TomShen1234:
@auspicacious The latest commit did add your last comments for the `enabled` docstring.

sphinx/source/index.rst
```diff
@@ -13,3 +13,22 @@ PnPQ
    api/modules
 
 PnPQ is a control library for hardware commonly used in quantum optical experiments (although it is probably useful for classical optical experiments as well!).
+
+Supported devices
+-----------------
+
+**ThorLabs**:
```
auspicacious, line 20: Continue to use semantic markup for headers, even in nested sections such as this. Avoid literal `:` (colons) in headers.

```suggestion
ThorLabs
^^^^^^^^
```

See https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

sphinx/source/index.rst
```diff
@@ -13,3 +13,22 @@ PnPQ
...
+
+- :py:mod:`Waveplates (kbd101) <pnpq.devices.waveplate_thorlabs_kb10crm>`
+- :py:mod:`Optical Delay Line (kbd101) <pnpq.devices.odl_thorlabs_kbd101>`
+- :py:mod:`Optical Switch (osw1310e) <pnpq.devices.switch_thorlabs_osw1310e>`
+- :py:mod:`Motorized Polarization Controller (mpc320) <pnpq.devices.polarization_controller_thorlabs_mpc>`
+
+**OzOptics**:
```
auspicacious, line 27:
```suggestion
OzOptics
^^^^^^^^
```

src/pnpq/devices/polarization_controller_thorlabs_mpc.py
```diff
@@ -46,21 +46,45 @@ class PolarizationControllerParams(TypedDict):
...
+            one for each channel.
+
+        """
 
     @abstractmethod
     def get_status(
         self, chan_ident: ChanIdent
     ) -> AptMessage_MGMSG_MOT_GET_USTATUSUPDATE:
-        pass
+        """Fetch the latest status of the device.
```
auspicacious, line 61:
```suggestion
        """Fetch the status of a single channel.
```

src/pnpq/devices/polarization_controller_thorlabs_mpc.py
```diff
@@ -46,21 +46,45 @@ class PolarizationControllerParams(TypedDict):
...
+        :param chan_ident: The motor channel to fetch status for.
+        :return: The message returned by the device, in
+            :py:class:`AptMessage_MGMSG_MOT_GET_USTATUSUPDATE`
+
+        """
 
     @abstractmethod
     def home(self, chan_ident: ChanIdent) -> None:
-        pass
+        """Move the device to home position
```
auspicacious, line 71:
```suggestion
        """Move the device to home position.
```

auspicacious (commented):
I think you should be able to just batch these suggestions and then we can call this PR done.

TomShen1234:
Done

auspicacious (approved):

Issue #120, opened 2025-04-18
Update ThorLabs Optical Delay Line to use APT protocol


TomShen1234:
Now that the Waveplate and MPC are using the `APTConnection` class, we should update the ODL to do the same. 

## Acceptance Criteria

- ODL Updated in the same way as the Waveplate and MPC
  - Add any messages necessary for basic operation
  - Adopt the APT Connection class
  - Abstract out public functions into an ABC
  - Update stub device
  - Add documentations
- The parent `OpticalDelayLine` class should be removed (at least for the ThorLabs part)
- Solves issue #43

auspicacious:
Sub-issues are all complete, closing.

PR #126, opened 2025-05-16
Refactors the ODL Driver

TomShen1234:
Implementation for #120
(Although I think a bit more should be added before it's closed)

Implemented the following:
- The driver itself (with ABC)
- Stub device
- Units
- Hardware test
- Unit test
- Modified MOVE_STOPPED message to support interruption of movements
- Updated timeout to better support longer homing of ODL

### DIFF
--- a/src/pnpq/devices/refactored_odl_thorlabs_kbd101.py
+++ b/src/pnpq/devices/refactored_odl_thorlabs_kbd101.py
@@ -134,10 +134,10 @@ class OpticalDelayLineThorlabsKBD101(AbstractOpticalDelayLineThorlabsKBD101):
                     # should decrease this interval.
                     self.connection.tx_ordered_sender_awaiting_reply.wait(0.9)
 
-    def identify(self, chan_ident: ChanIdent) -> None:
+    def identify(self) -> None:
         self.connection.send_message_no_reply(
             AptMessage_MGMSG_MOD_IDENTIFY(
-                chan_ident=chan_ident,
+                chan_ident=ChanIdent.CHANNEL_1,
                 destination=Address.GENERIC_USB,
                 source=Address.HOST_CONTROLLER,
             )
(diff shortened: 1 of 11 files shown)


src/pnpq/devices/refactored_odl_thorlabs_kbd101.py
```diff
@@ -0,0 +1,320 @@
...
+                )
+                and message.chan_ident == self._chan_ident
+                and message.destination == Address.HOST_CONTROLLER
+                and message.source == Address.GENERIC_USB
+            ),
+        )
+        # Sometimes the move stopped is received when interrupted
+        # by the user or when an invalid position is given
+        if isinstance(result, AptMessage_MGMSG_MOT_MOVE_STOPPED_20_BYTES):
+            self.log.warning(
```
auspicacious, line 201:
```suggestion
            self.log.error(
```

src/pnpq/devices/refactored_odl_thorlabs_kbd101.py
```diff
@@ -0,0 +1,320 @@
...
+                        message.position > absolute_distance - 1000
+                        and message.position < absolute_distance + 1000
+                    )
+                )
+            ),
+        )
+        # Sometimes the move stopped is received when interrupted
+        # by the user or when an invalid position is given
+        if isinstance(result, AptMessage_MGMSG_MOT_MOVE_STOPPED_20_BYTES):
+            self.log.warning(
```
auspicacious, line 247:
```suggestion
            self.log.error(
```

src/pnpq/devices/refactored_odl_thorlabs_kbd101.py
```diff
@@ -0,0 +1,320 @@
...
+                and message.source == Address.GENERIC_USB
+                and (  # If move is completed, check if the position is within 1mm of the target
+                    isinstance(message, AptMessage_MGMSG_MOT_MOVE_STOPPED_20_BYTES)
+                    or (
+                        message.position > absolute_distance - 1000
+                        and message.position < absolute_distance + 1000
+                    )
+                )
+            ),
+        )
```
auspicacious, line 243: It might be best to move the position check outside of the lambda.

That way, we can raise an exception if the position isn't what we expect it to be.

I'm trying to think of a scenario when we could receive a MOVE_COMPLETED that isn't associated with the move command that we sent, but that shouldn't be possible.

src/pnpq/units.py
```diff
@@ -252,6 +253,87 @@ def to_quantity(
     ),
 )
 
+# Custom units for brushless DC controller (kbd101, with DDSM100 stage)
+# TODO: In the future, we would want this to support a wider sets of stages
```
auspicacious, line 257:
```suggestion
# TODO: In the future, we would want this to support other stages
```

auspicacious (approved):

PR #131, opened 2025-05-23
Switch thorlabs osw

### LINKED ISSUE #130: Simple rewrite of SwitchThorlabsOsw1310e driver
Subissue to #129, made to be easier for beginners.

## Acceptance criteria

* Function names are consistent: right now one function is called `bar_state`, and the other is called `cross`, with `no _state`. Make this consistent by simply calling them `bar` and `cross`.
* There is an abstract base class defined so that we can build stub implementations that have the same API.
* The class is redefined as a dataclass, similar to the other Thorlabs drivers.

remonnnokawa:
closes #130 
I rewrite a code for OSW.
Linter denied  using name "bar" for function. So I use "bar_state" alternatively.


src/pnpq/devices/refactored_switch_thorlabs_osw1310e.py
```diff
@@ -0,0 +1,65 @@
+#
+# Thorlanbs Oprical Switch 1x2 and 2x2 1310E driver
```
auspicacious, line 2: It looks like these spelling errors were copied from the old switch:

```suggestion
# Thorlabs Optical Switch 1x2 and 2x2 1310E driver
```

src/pnpq/devices/refactored_switch_thorlabs_osw1310e.py
```diff
@@ -0,0 +1,65 @@
...
+class AbstractOpticalSwitchThorlabs1310E(ABC):
+    @abstractmethod
+    def connect(self) -> None:
+        """Open a new connection."""
+
+    def bar_state(self) -> None:
+        """Set the switch to the bar state."""
+
+    def cross_state(self) -> None:
+        """Set the switch to the cross state."""
```
auspicacious, line 22: All three methods must be decorated with `@abstractmethod`.

```suggestion
    @abstractmethod
    def connect(self) -> None:
        """Open a new connection."""

    @abstractmethod
    def bar_state(self) -> None:
        """Set the switch to the bar state."""

    @abstractmethod
    def cross_state(self) -> None:
        """Set the switch to the cross state."""
```

auspicacious (commented):
@master-of-ppap @remonnnokawa Please read this part before the individual comments.

If a pylint rule forces you to change a design requirement, it's best to write down what rule was broken so I don't have to re-create the issue locally.

That being said, this is an opportunity to reconsider the API's design.

What if we had an API that looked like this?

```python
class State(Enum):
    BAR = 1 # this might still trigger the pylint error, we could append _STATE to these
    CROSS = 2

...

class OSW:
    def set_state(self, state: State) -> None:
        # set either state from the same method
```

This might seem more awkward at first. But it makes it easier for us to do things like create a configuration file that specifies the OSW state, and then use that configuration file to automatically set the OSW to that state at startup. Maybe we'll want to do that someday.

master-of-ppap:
I agree with your comments! I think an `Enum` that stores the state of the device is a better idea. And it falls in line with the other devices (with `ChatIdent` as enums and stuff).

I'll try to tell these to @remonnnokawa and get him to fix it. But I want to try avoid helping him too much. He must learn.

auspicacious (commented):
This is partially my fault. I forgot how complex the connection initialization code is.

However, if something is unclear, please ask questions.

Before this PR is ready to merge, this code must be tested.

Before we can test the code, we need to:

* Read the official Thorlabs documentation, and make sure that we are sending messages to the device correctly. We also need to make sure that we are receiving and processing any messages that the device sends to us. Ask Amin where to find the documentation.
* If the messages are complex, we might need to model the messages we send and receive from the device as objects. We did this for the APT protocol messages [in this file](https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/apt/protocol.py).
* Implement a method that tells us the current state of the device. When we test the device, we need to be able to check the current state.

Then, we can add a test to the `hardware_tests` directory.

PR #137, opened 2025-06-04
Clean up some class variables

auspicacious:
While doing another PR it occurred to me that it would be better if we avoided using class variables as much as possible, to prevent confusion.

I haven't touched the ones related to channel configuration, because I think that might interfere with some of the work you're doing. I did remove some unused variables. I think it might be a good idea to make the channel variables into instance variables, though.

I haven't touched anything in the protocol classes, since those are not plausibly reconfigurable and I doubt anyone will ever try to change those values.


TomShen1234 (approved):
Good point

Issue #121, opened 2025-04-18
Remove duplicate legacy code


TomShen1234:
The `legacy` branch has already been created. Therefore, as much legacy code as possible should be removed. This includes the legacy waveplate class and the (soon to be) legacy optical delay line class.

## Acceptance criteria

* Semantic Versioning policy clearly explained in documentation.
* Note added to README and documentation to explain purpose of `legacy` branch
* Legacy driver classes that have a replacement are removed.
* Legacy driver classes that do not have a replacement are clearly marked in the documentation, so that it is clear their API is not stable and could be replaced in the future.

PR #140, opened 2025-06-19
Remove legacy code for K10CR1 and KBD101

TomShen1234:
Make some progress towards #121

Note that not all codes are removed yet, specifically, the OzOptics ODL is still present (and some helper files that it depends on). 

AFAIK `error.py` is used by legacy codes only, so it should probably also be removed after everything is cleaned. Correct me if I'm wrong.


auspicacious (approved):
@TomShen1234 

Good thinking on updating `legacy` again. The best way to communicate the change is probably to make a pull request to testbed-node moving it to the legacy branch, so that we have an example `pyproject.toml` we can share with other projects.

`errors.py` does contain one exception that is still in use, `InvalidStateException` (but you would have discovered this when you tried to delete the file). We might want to rename it to `exceptions.py`, and of course we'll want to delete the other exceptions, but the file should stick around.

PR #147, opened 2025-07-23
Added timing simulation for stub devices

### LINKED ISSUE #135: Add precise operation time simulation to stub devices
Good simulator  behaves like a real environment. 

So, to make better simulator, I need to include precise time informations so that I can compare scheduling protocols or routing protocols.

Currently, Stub devices only output logs, but I need to have those to contain time information **according to the device setting**.
For example ODL, the **operating time of ODL is differed by the distance**. Operating time of Wave Plate is differed by the angles. Thus, it is needed to take operating time according to the device setting.

Therefore, we need to incorporate realistic delays into the stub device behavior to simulate real operation timing.

## Acceptance Criteria

- Time informations for all of the stub devices are required.
- As it is written above, Operation time must be differed by the device setting.
- To behave like the device is operated, we cannot use `time.sleep()` since we are using [`asyncio`](https://docs.python.org/3.13/library/asyncio.html) and if we use `time.sleep()`, it will block everything.

master-of-ppap:
Closes #135 

## Additions
* Added timing for all (currently implemented and applicable) stub devices methods. MPC320, KBD101, K10CR1 stubs are all supported. 
  * Makes use of `time.sleep()`, and calculates the amount of time to wait based on the velocity of the device.
* Added unit tests to ensure that the waited time is expected. Mocked the `time.sleep()` module using `unittest.mock`. 
* Improved driver classes' parameters functionality by implementing a "parameters" class that inherits from `UserDict`. 
  * Simplified most `get_params`-esque methods for each device using this class.
* Fixed small bug on conversion from `kbd101_velocity` to `mm / s`.
  * (the output was being rounded, which introduced many div by zero errors in my code.. It took me like 3 hours to figure out the culprit...)


master-of-ppap:
Thanks copilot..

master-of-ppap:
Thanks for the review. I have added the requested changes! 

It was supposed to be a small PR but we made many mini changes along the way... Probably should have made a PR right then and there. Haha.

src/pnpq/stub_util.py
```diff
@@ -18,12 +18,9 @@ def sleep_delta_position(
...
     # device-specific positions, and standard time units (seconds).
     #
-    # Therefore, we first convert the delta position and velocity to a common unit
-    # in order to calculate the time to sleep.
-    #
-    # We must split the logic based on the velocity units,
-    # as some devices use degrees (mpc320, k10cr1) and others use meters (kbd101).
-    if velocity.units in ("mpc320_velocity", "k10cr1_velocity"):
+    # Therefore, we will first try to convert the velocity and delta position to degrees or meters,
+    # and if that fails, we will raise an error.
```
auspicacious, line 22: Maybe rewrite this a little to reflect what's happening.

> Therefore, we will first try to convert spatial dimension into degrees, and if that fails, into meters. If that conversion fails, we will allow the `DimensionalityError` to propagate.

PR #151, opened 2025-08-08
Refactors optical switch

### LINKED ISSUE #129: Rewrite SwitchThorlabsOsw1310e driver
The current [Thorlabs OSW 1310e](https://github.com/moonshot-nagayama-pj/PnPQ/blob/main/src/pnpq/devices/switch_thorlabs_osw1310e.py) driver is partially functional, but missing the features necessary for use in an integrated testbed system. For instance, there is no way to query the current status of the switch.

There is also some inconsistency in the API, and the design doesn't match the expectations we've set for other Thorlabs devices.

## Acceptance criteria

* Function names are consistent: right now one function is called `bar_state`, and the other is called `cross`, with no `_state`. Make this consistent by simply calling them `bar` and `cross`.
* There is an abstract base class defined so that we can build stub implementations that have the same API.
* The class is redefined as a dataclass, similar to the other Thorlabs drivers.
* The class can be initialized using a serial number. Initialization by serial port is not necessary at this time.
* The class is a context manager.
* The `current_state` method is implemented.
* All methods block until a reply, if any, is received and parsed.
* Commands are defined as constants, rather than ad-hoc strings.
* Basic hardware tests are defined in `hardware_tests`.
* Basic documentation is written.
* Basic event logs are defined.

TomShen1234:
Close #129

### DIFF
--- a/src/pnpq/devices/switch_thorlabs_osw1310e.py
+++ b/src/pnpq/devices/switch_thorlabs_osw1310e.py
@@ -1,48 +1,185 @@
-#
-# Thorlanbs Oprical Switch 1x2 and 2x2 1310E driver
-#       OSW12-1310E & OSW22-1310E
-#
+import time
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from enum import Enum
+from threading import Lock
+from types import TracebackType
+
 import serial
 import serial.tools.list_ports
 from serial import Serial
 
+from .utils import timeout
 
-class Switch:
-    def __init__(
+SET_STATE_BAR_COMMAND = b"S 1\x0a"
+SET_STATE_CROSS_COMMAND = b"S 2\x0a"
+GET_STATUS_COMMAND = b"S?\x0a"
+GET_QUERY_TYPE_COMMAND = b"T?\x0a"
+GET_BOARD_NAME_COMMAND = b"I?\x0a"
+
+
+class State(Enum):
+    BAR = 1
+    CROSS = 2
+
+
+class AbstractOpticalSwitchThorlabs1310E(ABC):
+    @abstractmethod
+    def set_state(self, state: State) -> None:
+        """Set the switch to the specified state."""
+
+    @abstractmethod
+    def get_status(self) -> State:
+        """Get the current state of the switch."""
+
+    # Get system information
+    @abstractmethod
+    def get_query_type(self) -> str:
+        """Get the query type of the switch."""
+
+    @abstractmethod
+    def get_board_name(self) -> str:
+        """Get the board name of the switch."""
+
+    @abstractmethod
+    def open(self) -> None:
+        """Open the serial connection to the switch."""
+
+    @abstractmethod
+    def close(self) -> None:
+        """Close the serial connection to the switch."""
+
+
+@dataclass(frozen=True, kw_only=True)
+class OpticalSwitchThorlabs1310E(AbstractOpticalSwitchThorlabs1310E):
+    # Required
+    serial_number: str
+
+    # All other properties are optional.
+
+    # Serial connection parameters. These defaults are used by all
+    # known Thorlabs devices that implement the APT protocol and do
+    # not need to be changed.
+    baudrate: int = field(default=115200)
+    bytesize: int = field(default=serial.EIGHTBITS)
+    exclusive: bool = field(default=True)
+    parity: str = field(default=serial.PARITY_NONE)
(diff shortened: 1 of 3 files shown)


src/pnpq/devices/switch_thorlabs_osw1310e.py
```diff
@@ -1,48 +1,185 @@
...
+                self.connection.flush()
+                self.connection.close()
+
+    def set_state(self, state: State) -> None:
+        with self.mutex_lock, timeout(3) as check_timeout:
+            while check_timeout():
+                if state == State.BAR:
+                    self.connection.write(SET_STATE_BAR_COMMAND)
+                elif state == State.CROSS:
+                    self.connection.write(SET_STATE_CROSS_COMMAND)
```
auspicacious, line 149: I don't think this is what you want to do. You want to write the command once, then check for its status. You don't want to keep writing the command over and over.

src/pnpq/devices/switch_thorlabs_osw1310e.py
```diff
@@ -1,48 +1,182 @@
...
+
+class State(Enum):
+    BAR = 1
+    CROSS = 2
+
+
+class AbstractOpticalSwitchThorlabs1310E(ABC):
+    @abstractmethod
+    def set_state(self, state: State) -> None:
+        """Set the switch to the specified state."""
```
auspicacious, line 23: These documentation comments are not very helpful. It's very frustrating when you are trying to learn how an API works and the documentation wastes your time.

There should be:

1. Class-level documentation stating that this is a thread-safe, blocking API
2. Class-level documentation naming the device(s) this is intended to support and, preferably, linking to the official product page
3. Method-level documentation for this function stating that it is [idempotent](https://en.wikipedia.org/wiki/Idempotence); if the switch is already in the desired state, setting it to the desired state again will not cause an error.

src/pnpq/devices/switch_thorlabs_osw1310e.py
```diff
@@ -1,48 +1,182 @@
...
+    def get_board_name(self) -> str:
+        """Get the board name of the switch."""
+
+    @abstractmethod
+    def open(self) -> None:
+        """Open the serial connection to the switch."""
+
+    @abstractmethod
+    def close(self) -> None:
+        """Close the serial connection to the switch."""
```
auspicacious, line 44: I'm not sure if these methods should be part of the API contract or not. On the one hand, it's easily possible to implement an optical switch that doesn't use the context-manager paradigm for opening and closing. On the other hand, it would make things easier if we can assume that all stub implementations can be used as context managers.

So, that being said, if we're going to make `open` and `close` part of the API contract, we should make `__enter__` and `__exit__` part of it as well.

auspicacious (commented):
Here's some improvements you can make while you don't have access to the device to re-test.

src/pnpq/devices/switch_thorlabs_osw1310e.py
TomShen1234: It came to be that although the device class is called the Thorlabs1310E, it appears that the same [manual](https://www.thorlabs.com/drawings/999104b27a1b1d36-4FE979AB-01C9-0476-6EA199A5090554BD/OSW22-1310E-Manual.pdf) applies to multiple switches models. Shall we adjust the naming slightly?

src/pnpq/devices/switch_thorlabs_osw1310e.py
```diff
@@ -1,48 +1,185 @@
...
+        raise ValueError(f"Unknown response: {response.decode('utf-8')}")
 
-    # def current_state(self):
-    #    if self.conn.is_open:
-    #        self.conn.write(b'S ?\x0A')
+    def get_query_type(self) -> str:
+        with self.mutex_lock:
+            self.connection.write(GET_QUERY_TYPE_COMMAND)
+            response = self.__read_serial_response()
+            return response.decode("utf-8")
```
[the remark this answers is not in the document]
TomShen1234: This seems to be an integer. However, it seems to always return 0 so I'm not sure what else is there. We can consider switching it to an integer.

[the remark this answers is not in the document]
auspicacious: No, we can leave it as a string unless you can find better information in the official documentation about this.

TomShen1234:
I'm thinking of doing the `State` enum modification in a later PR after this is done. It may require some changes to the way the enum is structured and how we test it.

PR #159, opened 2025-08-19
Finish up ODL mock

### LINKED ISSUE #134: Add mock driver for KBD101
Like the mock waveplate driver implemented in #103, we should implement the same for the KBD101 as well. This should be relatively trivial to do since the messages are the same as the K10CR1. It is also worth taking a look at whether certain parts of the mock driver can be abstracted and shared. 

## Acceptance Criteria
- A mock for `OpticalDelayLineThorlabsKBD101` created.
- It should accept the `AptMessage_MGMSG_MOT_MOVE_ABSOLUTE` and match the correct `AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES` as would have been sent from the real device.
- Unit test should be created for the `move_absolute` function of the driver.
- Assert that `send_message_expect_reply` was called the expected number of times.

master-of-ppap:
Closes #134 

## Addition
* Added all of the methods for the KBD odl mock


auspicacious:
I was right about the Pint unit issue in home params: there was a problem with the corresponding test populating a message with quantities expressed as Pint values rather than plain ints. I'm still not sure why mypy allowed us to set a Quantity on an int field, though: as far as I can tell, Quantity is not a subclass of int.

Regardless, that is fixed. I added a couple more tests for `set_channel_enabled`, which was the only public function of note with no tests. I also refactored some of the common asserts for the device's init conditions into a single method rather than replaying them everywhere; it would probably be better if we simply dropped these asserts entirely and had a separate test for the init process instead, but this is good enough for now and gives us a lot of feedback if we do something out of order in a test, I suppose.

I think we have at least happy-path coverage for everything now. I am going to merge this PR.

PR #168, opened 2025-09-03
Thorlabs OSWxx-yyyyE switch driver improvements

auspicacious:
* The driver's open/close behavior is now similar to the APT connection class: attempting to call methods while the object is closed raises an `InvalidStateException`, and, once closed, the object cannot be re-opened.
* The `get_query_type()` method has been renamed to `get_type_code()`; the original name was confusing the use of the term "query" in the official documentation, which called this function "query type code." This is technically a breaking change, but I don't think we need to be concerned about bumping the version number for this one as no-one is using it.
* Documentation improvements.
* I noticed that `hardware_tests` needed to be reorganized in the same way I reorganized `tests` in https://github.com/moonshot-nagayama-pj/PnPQ/pull/165 -- this should have been a separate pull request, but I did it here.

I think this covers everything I mentioned in [this comment](https://github.com/moonshot-nagayama-pj/PnPQ/issues/129#issuecomment-3235843062).

@TomShen1234 please have a look; I'll also have to test on real hardware before merging.


TomShen1234 (approved):
Ran the hardware test. Works fine.

PR #174, opened 2025-11-06
Fix support for K10CR2

driedbones05:
Close #174
Since I’ve already run the bash file and it completed smoothly, I’m not sure whether there are any issues in the code. Here’s what I did:

- Added a new subclass with a different response structure to handle 34-byte messages.

- Modified existing classes and added `AptMessage_MGMSG_MOT_GET_STATUSUPDATE_34_BYTES`, `AptMessage_MGMSG_MOT_MOVE_COMPLETED_34_BYTES`, and `AptMessageWithDataMotorStatus_34_BYTES`.

- Updated the test files to accommodate the new parameters.

### DIFF
--- a/src/pnpq/apt/protocol.py
+++ b/src/pnpq/apt/protocol.py
@@ -1370,7 +1370,7 @@ class AptMessage_MGMSG_MOT_MOVE_COMPLETED(AptMessage):
         if length == 20:
             return AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES.from_bytes(raw)
         raise ValueError(
-            f"Expected data packet length 6 or 20 or 34, but received {length} instead. Full raw data was {raw!r}"
+            f"Expected data packet length 6 or 20, but received {length} instead. Full raw data was {raw!r}"
         )
 
 
@@ -1520,7 +1520,7 @@ class AptMessage_MGMSG_MOT_MOVE_STOPPED(AptMessage):
         if length == 20:
             return AptMessage_MGMSG_MOT_MOVE_STOPPED_20_BYTES.from_bytes(raw)
         raise ValueError(
-            f"Expected data packet length 6 or 20 or 34, but received {length} instead. Full raw data was {raw!r}"
+            f"Expected data packet length 6 or 20, but received {length} instead. Full raw data was {raw!r}"
         )
 
 
--- a/tests/pnpq/apt/test_protocol.py
+++ b/tests/pnpq/apt/test_protocol.py
@@ -369,7 +369,7 @@ def test_AptMessage_MGMSG_MOT_GET_STATUSUPDATE_34_bytes_to_bytes() -> None:
         reserved3=0x00000000,
     )
     assert msg.to_bytes() == bytes.fromhex(
-        "8104 1c00 81 22 0100 01000000 00000000 07000000 0100 00000000 00000000 00000000"  # Temporarily Fix
+        "8104 1c00 81 22 0100 01000000 00000000 07000000 0100 00000000 00000000 00000000"
     )
 
 
--- a/tests/pnpq/devices/test_waveplate_thorlabs_k10cr1.py
+++ b/tests/pnpq/devices/test_waveplate_thorlabs_k10cr1.py
@@ -34,7 +34,7 @@ from pnpq.devices.waveplate_thorlabs_k10cr1 import (
 from pnpq.errors import InvalidStateException
 from pnpq.units import pnpq_ureg
 
-ustatus_message = AptMessage_MGMSG_MOT_GET_STATUSUPDATE_20_BYTES(
+status_message = AptMessage_MGMSG_MOT_GET_STATUSUPDATE_20_BYTES(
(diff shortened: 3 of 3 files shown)


src/pnpq/devices/waveplate_thorlabs_k10cr1.py
```diff
@@ -331,16 +334,26 @@ def move_absolute(self, position: Quantity) -> None:
...
+                    isinstance(message, AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES)
+                    and message.chan_ident == self._chan_ident
+                    and message.position == absolute_distance
+                    and message.destination == Address.HOST_CONTROLLER
+                    and message.source == Address.GENERIC_USB
+                )
+                or (
+                    isinstance(message, AptMessage_MGMSG_MOT_MOVE_COMPLETED_34_BYTES)
+                    and message.chan_ident_1 == self._chan_ident
+                    and message.chan_ident_2 == self._chan_ident
```
auspicacious: I don't know if we can assume that `chan_ident_2` is also set to the real chan ident, and we really don't care -- we shouldn't try to match on it here.
```suggestion
```

tests/pnpq/apt/test_protocol.py
```diff
@@ -336,6 +339,42 @@ def test_AptMessage_MGMSG_MOT_GET_STATUSUPDATE_to_bytes() -> None:
...
+        chan_ident_2=ChanIdent.CHANNEL_1,
+        position=1,
+        enc_count=0,
+        status=Status(CWHARDLIMIT=True, CCWHARDLIMIT=True, CWSOFTLIMIT=True),
+        reserved1=0x00000000,
+        reserved2=0x00000000,
+        reserved3=0x00000000,
+    )
+    assert msg.to_bytes() == bytes.fromhex(
+        "8104 1c00 81 22 0100 01000000 00000000 07000000 0100 00000000 00000000 00000000"  # Temporarily Fix
```
auspicacious: "Temporarily fix?" I'm not sure what that means or what is being fixed.

tests/pnpq/devices/test_waveplate_thorlabs_k10cr1.py
```diff
@@ -33,14 +34,26 @@
 from pnpq.errors import InvalidStateException
 from pnpq.units import pnpq_ureg
 
-ustatus_message = AptMessage_MGMSG_MOT_GET_STATUSUPDATE(
+ustatus_message = AptMessage_MGMSG_MOT_GET_STATUSUPDATE_20_BYTES(
```
auspicacious: This is not your fault, but we should probably rename this variable, since this is not a ustatus message.

auspicacious (commented):
I think that this code will _almost_ work when we test it with the real device. I think the only changes that are necessary to make it work are to remove the check

```
and message.chan_ident_2 == self._chan_ident
```

You have also written more code than is necessary. We don't know if there is a 34-byte version of messages like `MOVE_STOP`. We will need to test and find out. I would like to know where the format that you implemented here came from.

Let's make that one fix and test it.

src/pnpq/apt/protocol.py
```diff
@@ -656,6 +656,90 @@ def to_bytes(self) -> bytes:
         )
 
 
+@dataclass(frozen=True, kw_only=True)
+class AptMessageWithDataMotorStatus_34_BYTES(AptMessageWithData):
```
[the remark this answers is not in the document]
driedbones05: Oh, regarding to this one. There is no version of this message, I only added because I thought `AptMessage_MGMSG_MOT_GET_STATUSUPDATE_34_BYTES` will not able to use the 20 bytes message struct so I created a new class for it.

driedbones05:
After what you commented, I did realized that these classes `AptMessage_MGMSG_MOT_MOVE_COMPLETED_34_BYTES` `AptMessage_MGMSG_MOT_MOVE_STOPPED_34_BYTES` not suppose to be created since the document doesn't appear to adapt 34 bytes messages

tests/pnpq/devices/test_waveplate_thorlabs_k10cr1.py
```diff
@@ -33,14 +34,26 @@
...
+    chan_ident_2=ChanIdent(2),
+    destination=Address.HOST_CONTROLLER,
+    source=Address.GENERIC_USB,
+    enc_count=0,
+    position=0,
+    status=Status(INMOTIONCCW=True, INMOTIONCW=True, HOMED=True),
+    reserved1=0,
+    reserved2=0,
+    reserved3=0,
+)
```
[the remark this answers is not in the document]
driedbones05: I'll try to think how to use this data since I have no idea what to do with it at the moment.

auspicacious:
@driedbones05 I'm testing with the real hardware right now and it looks like your fix is working -- however there's another problem with a different message. I'm going to make some changes directly to your branch to fix it.

auspicacious (approved):
@driedbones05 I'm going to merge this and we're going to make a new release. There's a few things that we need to follow up on. I will create another issue for those.

PR #179, opened 2025-11-19
Fuzzy position matching for K10CR move_absolute, better connection error handling

### LINKED ISSUE #173: Log errors on connection thread shutdown
The `apt.connection` module implements a connection class which encapsulates the complexity of talking with a Thorlabs device over a USB connection.

There are two threads which are created when `.open()` is called on an instance of this class:

* The dispatcher thread: this receives messages from the device and translates them into message objects
* The ordered sender thread: this receives messages from a queue and sends them to the device

Shutting down the connection requires shutting down these threads. Moreover, if there is an error in either thread, the entire connection should shut down.

This all works correctly. However, if one of the threads experiences an error, that error is not being logged. This makes connection problems much more difficult to debug.

The challenge is that "normal" shutdown also causes errors. We don't want to log those.

## Acceptance criteria

* Errors in the connection threads are logged at level `ERROR`.
* "Normal" errors caused by `.close()` are not logged.

auspicacious:
Resolves https://github.com/moonshot-nagayama-pj/PnPQ/issues/173

Based on user reports, it seems like the K10CR2 (and possibly the CR1) sometimes overshoots position 0 when `move_absolute(0)` is called. When it bounces back to the zero-position, it may still register as a few steps away from 0; we saw `MOVE_COMPLETED` messages with a position of -1. As there are are 136,533 steps per each degree of movement, -1 is still extremely close to 0.

This PR adds fuzzy matching, so that anything within +-0.1 degrees of the desired position is considered valid. This way, we can still reject wildly inaccurate moves and eventually signal them as errors.

For a while now (see https://github.com/moonshot-nagayama-pj/PnPQ/issues/173), we have been meaning to fix connection error logging to help make debugging issues like this easier, as some exceptions were disappearing when threads exited; I've added that to this PR as well. Finally, I increased the receive wait timeout from 5 to 6 minutes and made the timeout's behavior easier to understand.

### DIFF
--- a/hardware_tests/conftest.py
+++ b/hardware_tests/conftest.py
@@ -17,7 +17,7 @@ def excepthook(
     e: BaseException,
     traceback: TracebackType | None,
 ) -> Any:
-    log.error(event=Event.UNCAUGHT_EXCEPTION, exc_info=e)
+    log.error(event=Event.UNEXPECTED_EXCEPTION, location="excepthook", exc_info=e)
     return sys.__excepthook__(exception_type, e, traceback)
 
 
@@ -28,7 +28,11 @@ original_threading_excepthook = threading.excepthook
 
 
 def threading_excepthook(args: Any) -> Any:
-    log.error(event=Event.UNCAUGHT_EXCEPTION, exc_info=args.exc_value)
+    log.error(
+        event=Event.UNEXPECTED_EXCEPTION,
+        location="threading_excepthook",
+        exc_info=args.exc_value,
+    )
     return original_threading_excepthook(args)
 
 
--- a/hardware_tests/pnpq/devices/test_waveplate_thorlabs_k10cr1.py
+++ b/hardware_tests/pnpq/devices/test_waveplate_thorlabs_k10cr1.py
@@ -17,7 +17,7 @@ from pnpq.units import pnpq_ureg
 
 log = structlog.get_logger()
 
-SERIAL_NUMBER = "55409764"
+SERIAL_NUMBER = "55528654"
 
 
 def test_connection() -> None:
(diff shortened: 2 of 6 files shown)


auspicacious:
Waiting to test with real hardware before merging and creating a patch release.

auspicacious:
Test passed on the hardware, I even was able to validate that in the logs we were seeing position -1:

```
{"message": "AptMessage_MGMSG_MOT_MOVE_COMPLETED_20_BYTES(destination=<Address.HOST_CONTROLLER: 1>, source=<Address.GENERIC_USB: 80>, chan_ident=<ChanIdent.CHANNEL_1: 1>, position=-1, velocity=0, motor_current=<Quantity(0, 'milliampere')>, status=UStatus(CWHARDLIMIT=False, CCWHARDLIMIT=False, CWSOFTLIMIT=False, CCWSOFTLIMIT=False, INMOTIONCW=False, INMOTIONCCW=False, JOGGINGCW=False, JOGGINGCCW=False, CONNECTED=False, HOMING=False, HOMED=True, INITILIZING=False, TRACKING=False, SETTLED=False, POSITIONERROR=False, INSTRERROR=False, INTERLOCK=False, OVERTEMP=False, BUSVOLTFAULT=False, COMMUTATIONERROR=False, DIGIP1=False, DIGIP2=False, DIGIP3=False, DIGIP4=False, OVERLOAD=False, ENCODERFAULT=False, OVERCURRENT=False, BUSCURRENTFAULT=False, POWEROK=False, ACTIVE=False, ERROR=False, ENABLED=True))", "event": "rx_message_known", "level": "debug", "timestamp": "2025-11-21T02:37:22.498102Z"}
```

PR #182, opened 2025-12-17
add new exception class for thorlabs_osw

### LINKED ISSUE #169: Handle Thorlabs OSWxx-yyyyE command error responses
Currently, we assume that all commands sent to Thorlabs optical switches are successful. However, the device may respond to a command with an error message, as detailed in the official documentation (pages 10 and 11):

> The OSW12(22)-xxxE board responds to command errors with an error message containing an error code and a description.
>
> Syntax: `Error nn,Descriptive text<CR><LF>` (where nn is the error code).

The error codes are defined as:

```
Error Codes
Error code Description
01 A general system error occurred
02 A math domain error was detected
03 The given value is out of range
06 Non-volatile memory error
10 A communication error occurred
11 The command is unknown
12 Wrong number of command parameters
13 The command parameter is invalid
```

We should handle these errors if they occur by translating them into an exception, preferably one which preserves the original error code and text.

## Acceptance criteria
 
* All command implementations translate error replies into an exception which contains the error code and descriptive text. It is not necessary to distinguish between errors; all should be considered fatal.
* Manually tested on real device by, for instance, temporarily modifying the driver to send an out-of-range value or an unknown command.
* Automated testing added using a mock serial connection to send an error response.
* Documentation added to all methods that may throw this exception.
* The mock driver does not need to support error generation at this time.

driedbones05:
Close #169 

I temporarily finished this task (for now), I got some problem that I cannot test the code on locally on my computer (sorry to all the people who got disturbed by status update messages)
I want to implement an error to test if this work or not but what I added and changes are:
- Create a new class to create exception raised for errors reported by Thorlabs OSW devices and add a new parser 
- Changes some of the function so that they able to raise exception. 

There is point that I haven't done by now (will try it but not really sure) : 

> Manually tested on real device by, for instance, temporarily modifying the driver to send an out-of-range value or an unknown command.

### DIFF
--- a/src/pnpq/devices/switch_thorlabs_osw_e.py
+++ b/src/pnpq/devices/switch_thorlabs_osw_e.py
@@ -250,3 +250,36 @@ class OpticalSwitchThorlabsE(AbstractOpticalSwitchThorlabsE):
     def _fail_if_closed(self) -> None:
         if (not self._opened_event.is_set()) or self._closed_event.is_set():
             raise InvalidStateException("Tried to use a closed switch object.")
+class ThorlabsOswError(Exception):
+    """Raised when a Thorlabs OSWxx-yyyyE optical switch reports an error
+    or sends an invalid/unexpected response.
+
+    Attributes
+    ----------
+    code : int | None
+        The numeric error code from the device (e.g. 1, 3, 11), or None if
+        the reply could not be parsed.
+        The codes correspond to the Thorlabs manual, e.g.:
+          01: A general system error occurred
+          02: A math domain error was detected
+          03: The given value is out of range
+          06: Non-volatile memory error
+          10: A communication error occurred
+          11: The command is unknown
+          12: Wrong number of command parameters
+          13: The command parameter is invalid
+    raw_reply : str
+        The raw reply line from the device.
+    """
+
+    def __init__(self, code, description: str, raw_reply: str) -> None:
+        self.code = code
+        self.description = description
+        self.raw_reply = raw_reply
+
+        if isinstance(code, int):
+            code_str = f"{code:02d}"
+        else:
+            code_str = "Unknown Code"
+
+        super().__init__(f"Thorlabs OSW error {code_str}: {description}\nRaw reply: {raw_reply}")
--- a/src/pnpq/devices/switch_thorlabs_osw_e.py
+++ b/src/pnpq/devices/switch_thorlabs_osw_e.py
@@ -28,6 +28,10 @@ class AbstractOpticalSwitchThorlabsE(ABC):
         This function is idempotent; if the switch is already in the desired state, setting it to the same state again will not cause an error.
 
         :param state: The state to set the switch to.
+
+        :raises InvalidStateException: If the switch object is closed.
+        :raises ThorlabsOswError: If the device reports an error or return
+            an unexpected reply while setting or checking the state.
         """
 
     @abstractmethod
@@ -35,6 +39,9 @@ class AbstractOpticalSwitchThorlabsE(ABC):
         """Get the current state of the switch.
 
         :return: The current state of the switch.
+        :raises InvalidStateException: If the switch object is closed.
+        :raises ThorlabsOswError: If the device reports an error or return
+            an unexpected reply while setting or checking the state.
         """
 
     # Get system information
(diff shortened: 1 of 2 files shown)


src/pnpq/devices/switch_thorlabs_osw_e.py
```diff
@@ -213,32 +224,103 @@ def get_state(self) -> State:
...
+        """
         command = b"S?\n"
         self._connection.write(command)
         response = self._read_serial_response()
-        return State(int(response.decode("utf-8")))
+        try:
+            value_str = response.decode("utf-8")
+            value_int = int(value_str)
+            return State(value_int)
+        except Exception as e:
```
auspicacious, line 238: The linter should be warning you that you shoudn't catch `Exception`, I'm not sure why it's passing. There is no reason to catch `Exception` here. The original exceptions from `decode` and `int` and `State` are just as useful as the wrapped `ThorlabsOswError`.

src/pnpq/devices/switch_thorlabs_osw_e.py
```diff
@@ -213,32 +224,103 @@ def get_state(self) -> State:
...
+        :raises ThorlabsOswError: If the device reports an error or sends an unexpected reply that cannot be parsed as a valid state.
+        """
         command = b"S?\n"
         self._connection.write(command)
         response = self._read_serial_response()
-        return State(int(response.decode("utf-8")))
+        try:
+            value_str = response.decode("utf-8")
+            value_int = int(value_str)
+            return State(value_int)
```
auspicacious, line 237: I'm not sure why we've taken a single line of code and rewritten it as three lines.

src/pnpq/devices/switch_thorlabs_osw_e.py
```diff
@@ -213,32 +224,103 @@ def get_state(self) -> State:
...
+        except Exception as e:
+            # This help to catch decoding erros and integer issues
+            raise ThorlabsOswError(
+                code=None,
+                description="Received invalid state response from Thorlabs OSW device.",
+                raw_reply=response.decode("utf-8", errors="replace"),
+            ) from e
```
auspicacious, line 244: This doesn't make a lot of sense. The idea of the `ThorlabsOswError` is that it reports error codes received from the device. There is no error code being sent from the device in this situation.

src/pnpq/devices/switch_thorlabs_osw_e.py
```diff
@@ -213,32 +224,103 @@ def get_state(self) -> State:
...
             response = self._read_serial_response()
-            return response.decode("utf-8")
+            try:
+                return response.decode("utf-8")
+            except UnicodeDecodeError as e:
+                raise ThorlabsOswError(
+                    code=None,
+                    description="Received non-decodable type code from Thorlabs OSW device.",
+                    raw_reply=response.decode("utf-8", errors="replace"),
+                ) from e
```
auspicacious, line 264: If you're worried about decoding errors, there's a better way to handle them: don't use `utf-8`. We could think of that as a bug in the application. Use a single-byte encoding scheme instead. Specifically, use `iso-8859-1`.

https://docs.python.org/3/library/codecs.html#module-encodings

Using utf-8 here was unlikely to fail (I'm not expecting characters outside of the basic 128 ASCII values) but using a single-byte encoding will remove the possibility of failure entirely, I think.

Also, dealing with this kind of error was not part of the assigned issue.

src/pnpq/devices/switch_thorlabs_osw_e.py
```diff
@@ -213,32 +224,103 @@ def get_state(self) -> State:
...
+        if not response:
+            raise ThorlabsOswError(
+                code=None,
+                description="No response received from Thorlabs OSW device.",
+                raw_reply="",
+            )
+        if response.endswith(b"\r\n"):
+            decoding_messages = response[:-2]
+        else:
+            decoding_messages = response
```
auspicacious, line 309: I can understand why you might want to handle partial responses like this, but it is outside the scope of this issue.

src/pnpq/devices/switch_thorlabs_osw_e.py
```diff
@@ -213,32 +224,103 @@ def get_state(self) -> State:
...
+        # Decode to inspect for "Error.." messages
+        try:
+            text = decoding_messages.decode("utf-8")
+        except UnicodeDecodeError as e:
+            raise ThorlabsOswError(
+                code=None,
+                description="Received non-decodable response from Thorlabs OSW device.",
+                raw_reply=response.decode("utf-8", errors="replace"),
+            ) from e
+        stripped_text = text.strip()
```
auspicacious, line 320: It doesn't really make sense to strip the text here.

src/pnpq/devices/switch_thorlabs_osw_e.py
```diff
@@ -213,32 +224,103 @@ def get_state(self) -> State:
...
+        if stripped_text.lower().startswith("error"):
```
auspicacious, line 321: I don't think the specification says that the message format is case-insensitive.

src/pnpq/devices/switch_thorlabs_osw_e.py
```diff
@@ -213,32 +224,103 @@ def get_state(self) -> State:
...
+            raise parse_thorlabs_osw_error(stripped_text)
+        return decoding_messages
```
auspicacious, line 323: If we're already decoding to a string... shouldn't we just return the string here?

src/pnpq/errors.py
```diff
@@ -63,3 +66,83 @@ class InvalidStateException(Exception):
...
+class ThorlabsOswError(Exception):
+    """Exception raised for errors reported by Thorlabs OSWxx-yyyyE devices.
+
+    The instance provides three attributes: ``code`` (int or None),
+    ``description`` (str), and ``raw_reply`` (str).
+    """
+
+    def __init__(
+        self,
+        code: Optional[int],
```
auspicacious, line 80: There should be an open-ended enum that names the error codes. This shouldn't be an int.

auspicacious (commented):
This code does not match what was requested in the issue.

auspicacious:
Closed, we are focusing efforts elsewhere right now.

PR #184, opened 2026-04-22
Increase homing time_out K10CR1/M rotator cage

hhatpsi:
Update home() timeout in Thorlabs K10CR1/M rotator cages.

### DIFF
--- a/src/pnpq/devices/waveplate_thorlabs_kb10crm.py
+++ b/src/pnpq/devices/waveplate_thorlabs_kb10crm.py
@@ -66,7 +66,7 @@ class Waveplate:
 
         self.resolution = 136533
         self.max_steps = 136533
-        self.rotate_timeout = 30
+        self.rotate_timeout = 120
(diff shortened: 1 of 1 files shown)


src/pnpq/devices/waveplate_thorlabs_kb10crm.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         self.resolution = 136533
         self.max_steps = 136533
         self.rotate_timeout = 30
-        self.home_timeout = 30
+        self.home_timeout = 120
```
auspicacious, line 70: @hhatpsi I strongly recommend increasing the other timeouts in the file as well.

hhatpsi: @auspicacious applied for the rotate_timeout as well

hhatpsi (commented):
applied for the rotate_timeout as well

